PR #8, opened 2018-12-23
do some dce before printing

vchuravy:
Should get rid of annoying goto if not false bb left
over from boundschecking.

The tests don't work since tests force boundscheck...

CC: @yingboma


YingboMa:
Awesome! Thanks!

PR #10, opened 2019-01-25
Intelligently decide what to print in limited screen space

Keno:
Function calls are often too large for a single line, messing up the menu.
That's probably primarily a TerminalMenus bug (and it should be fixed to
support multi-line display), but even so in extreme cases (*cough* Zygote),
function calls can be multiple terminal pages. This implements some logic
to always print function calls as a single line, falling back through the
following progression:
- Full argument types
- Only the head datatype, but not their parameters
- Only the number of arguments
- Nothing

E.g.:
```
In `Zygote._forward` select a call to descend into or ↩ to ascend. [q] to quit.
   %5 = invoke Zygote._forward(…,…,…,…)
   %18 = invoke getindex(::Tuple,::Int64)::Union{}
   %33 = invoke getindex(::Tuple,::Int64)::Union{}
   %65 = invoke getindex(::Tuple,::Int64)::Union{}
   %80 = invoke getindex(::Tuple,::Int64)::Union{}
   %270 = invoke getindex(::Tuple,::Int64)::Union{}
```

@vchuravy

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -3,84 +3,12 @@ module Cthulhu
 using TerminalMenus
 using InteractiveUtils
 
+using Core: MethodInstance
+const Compiler = Core.Compiler
+
+include("callsite.jl")
+
 export descend, @descend, descend_code_typed, descend_code_warntype, @descend_code_typed, @descend_code_warntype
-
-struct Callsite
-    id::Int # ssa-id
-    f
-    tt
-    rt
-end
-
-mutable struct TextWidthLimiter
-    io::IO
-    width::Int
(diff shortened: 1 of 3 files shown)


Keno:
Actually, I have more changes to this, so hold on.

vchuravy:
Thanks! Only comment sofar is that this should go into a different source file. I try to only have the main logic in `Cthulhu.jl`

vchuravy:
Oh cool! Would have taken quite a while for me to reverse engineer that ;)

Keno:
Thanks. I actually have more fixes here. I'll put up a separate PR

PR #19, opened 2019-02-08
Pass `params` through descend calls and add option to dump the cache

vchuravy:
Very helpful for debugging Cassette issues (cc: @jrevels )

@Keno do you forsee any issue with re-use the params cache? I was thinking about creating a deepcopy before each descend so that one can go up, without potentially poisoning the cache.


Keno:
> @Keno do you forsee any issue with re-use the params cache? I was thinking about creating a deepcopy before each descend so that one can go up, without potentially poisoning the cache.

No, that seems ok. We may want to consider also disabling the global cache in addition, though that might require some Base work.

PR #21, opened 2019-02-12
Add special handling for return_type

Keno:
The idea is that the user will generally want to descend into
function that we're querying the return type of. However,
to differentiate this from a regular function call, I add a
special kind of callsite info to indicate this kind of call site
such that we may print it specially.


vchuravy:
Interesting, I agree that we should probably print those special. I have been wanting to add printing for filtered out-callsites.

LGTM.

PR #44, opened 2019-07-08
Request: Syntax-highlight for LLVM IR and native code

tkf:
Cthulhu is a very useful tool to look at generated LLVM IR and native code but one thing I'm missing is syntax highlighting.

So, I wrote this simple patch to use [`pygmentize`](http://pygments.org/docs/cmdline/) to highlight LLVM IR and ASM if it is available.  Is there any chance something like this can be merged to Cthulhu?

If hard-coding `pygmentize` usage in Cthulhu is not a good idea, one option would be to write a thin Julia wrapper for `pygmentize`. I think I can write one as I've been using my unregistered package [ColorfulCodeGen.jl](https://github.com/tkf/ColorfulCodeGen.jl) which has most of the code needed to do this.  Of course, it would be nice if there is a pure Julia solution.  But I couldn't find any generic code highlighter in Julia.

### DIFF
--- a/src/codeview.jl
+++ b/src/codeview.jl
@@ -2,8 +2,7 @@ const asm_syntax = Ref(:att)
 
 function highlight(io, lexer, x)
     Sys.which("pygmentize") === nothing && return print(io, x)
-    open(pipeline(`pygmentize -f terminal256 -O style=lovelace -l $lexer`;
-                  stdout=io, stderr=stderr), "w") do io
+    open(pipeline(`pygmentize -l $lexer`; stdout=io, stderr=stderr), "w") do io
         print(io, x)
     end
 end
--- a/src/codeview.jl
+++ b/src/codeview.jl
@@ -1,28 +1,57 @@
+Base.@kwdef mutable struct CthulhuConfig
+    enable_highlighter::Bool = false
+    highlighter::Cmd = `pygmentize -l`
+end
+
+highlighter_exists(config::CthulhuConfig) =
+    Sys.which(config.highlighter.exec[1]) !== nothing
+
+@init begin
+    CONFIG.enable_highlighter = highlighter_exists(CONFIG)
+end
+
+"""
+    Cthulhu.CONFIG
+
+# Options
+- `enable_highlighter::Bool`: Use command line `highlighter` to syntax highlight
+  LLVM and native code.  Set to `true` if `highlighter` exists at the import
+  time.
+- `highlighter::Cmd`: A command line program that receives "llvm" or "asm" as
+  an argument and the code as stdin.  Defaults to `$(CthulhuConfig().highlighter)`.
+"""
+const CONFIG = CthulhuConfig()
+
 const asm_syntax = Ref(:att)
 
-function highlight(io, lexer, x)
-    Sys.which("pygmentize") === nothing && return print(io, x)
-    open(pipeline(`pygmentize -l $lexer`; stdout=io, stderr=stderr), "w") do io
+function highlight(io, x, lexer, config::CthulhuConfig)
+    config.enable_highlighter || return print(io, x)
+    if !highlighter_exists(config)
+        @warn "Highlighter command $(config.highlighter.exec[1]) does not exist."
+        return print(io, x)
+    end
+    cmd = `$(config.highlighter) $lexer`
+    open(pipeline(cmd; stdout=io, stderr=stderr), "w") do io
         print(io, x)
     end
 end
 
-function cthulhu_llvm(io::IO, mi, optimize, debuginfo, params)
+function cthulhu_llvm(io::IO, mi, optimize, debuginfo, params, config::CthulhuConfig)
     dump = InteractiveUtils._dump_function_linfo(
         mi, params.world, #=native=# false,
         #=wrapper=# false, #=strip_ir_metadata=# true,
         #=dump_module=# false, #=syntax=# asm_syntax[],
         optimize, debuginfo ? :source : :none)
-    highlight(io, "llvm", dump)
+    highlight(io, dump, "llvm", config)
 end
 
-function cthulhu_native(io::IO, mi, optimize, debuginfo, params)
+function cthulhu_native(io::IO, mi, optimize, debuginfo, params, config::CthulhuConfig)
     dump = InteractiveUtils._dump_function_linfo(
         mi, params.world, #=native=# true,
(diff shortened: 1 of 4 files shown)


vchuravy:
Interesting! I am not a big fan of relying on an external program, but I think if we outline the check the a `deps/build.jl` generating an `build/ext.jl` (like the CUDA stack) which checks
if `pygments` or another highlighter is available, that would be fine.

A pure Julia solution would be cool!

src/codeview.jl
```diff
@@ -1,12 +1,20 @@
 const asm_syntax = Ref(:att)
 
+function highlight(io, lexer, x)
+    Sys.which("pygmentize") === nothing && return print(io, x)
```
vchuravy: This checks should be outline to a build step

src/codeview.jl
```diff
@@ -1,12 +1,20 @@
...
+    open(pipeline(`pygmentize -f terminal256 -O style=lovelace -l $lexer`;
```
vchuravy: We should probably detect if the terminal is able to handle that/have an option to turn it off

src/codeview.jl
tkf: Is outlining necessary for interactive tools like Cthulhu?  I understand that this is important when the existence and details of the external tools matter in the performance sensitive part (like CUDA).  But `Sys.which` is negligible compared to human interactions.  On the other hand, being able to use code highlighting right after installing `pygmentize` without needing to re-build/precompile seems to be a slight win.

src/codeview.jl
tkf: Do you know if there is a cross platform way to do this?  I can use `tput colors` in Linux but it also has an issue when used in `tmux` https://unix.stackexchange.com/q/118806. `REPL` has `REPL.GlobalOptions.hascolor` but it's just a boolean value.

I think the simplest solution would be to make it configurable.  It also avoids hard-coding color scheme `-O style=lovelace`.  I chose it because I found that it works nicely with dark and light background.  But other users may have other preference.

tkf: Actually, it looks like `pygmentize`'s default setup is OK.  It has some ad-hoc mechanism to detect the color setting (check if `$TERM` has `256` as a substring) https://bitbucket.org/birkenfeld/pygments-main/src/e63f7b68898445836091b51cfa81bf0613fb05eb/pygments/cmdline.py#lines-465:468

src/codeview.jl
vchuravy: I like efficiency ;), but as long as I can turn it off, when it breaks. Preferably it should be a toggle like optimized

tkf: I do like efficiency too.  My point is that it is highly inefficient to require users to restart entire Julia REPL to enable or disable highlighting.  (OK, if they use Revise.jl maybe that's fine.  But basing the interface design on to Revise.jl does not seem like a good idea to me.)

For the same reason, I don't think hiding it behind the toggling UI is a good idea.  The biggest bottleneck here is the human keystroke.  Saving single keystroke at the cost of single `Sys.which` is a huge win.

vchuravy: By putting it behind `Sys.which` you are forcing me to remove `pygments` to disable it...

1. Precompute during `__init__` and put the result into a global `Ref`.
2. There needs to be a toggle to turn it *off* at least, I don't want to restart a `descend` just because the highlighter broke

src/codeview.jl
vchuravy: Okay that sounds good.

src/codeview.jl
tkf: That's also the reason why I was suggesting to add a global configuration.  It didn't occurred to me that you'd want to toggle highlighter on-the-fly but it makes sense, too.

So how about something like this?

```julia
Base.@kwdef mutable struct CthulhuConfig
    enable_highlighter::Bool = false
    highlighter::Cmd = `pygmentize`
end

const CONFIG = CthulhuConfig()

function __init__()
    CONFIG.enable_highlighter = Sys.which(CONFIG.highlighter) !== nothing
end

function _descend(...)
    ...
        elseif toggle === :highlighter
            # Make the UI toggle "sticky"; i.e., the next descent will remember
            # the setting.
            CONFIG.enable_highlighter ⊻= true
    ...
end
```

The reason why I separated `enable_highlighter` and `highlighter` is that it lets you toggle highlighter without discarding user setting in `highlighter`.

This is similar to how ElectronDisplay.jl handles configuration: https://github.com/queryverse/ElectronDisplay.jl/blob/4d0c799017d1f1cfc7ee42d82040b3666e7b7f66/src/ElectronDisplay.jl#L43-L61

vchuravy: That looks good!

tkf:
OK, I updated the UI. Here is an example session (typing `s`, `s`, and `L`)

image

test/runtests.jl
```diff
@@ -101,3 +101,39 @@ let callsites = find_callsites_by_ftt(callf, Tuple{Union{typeof(sin), typeof(cos
...
+            @test_logs (:warn, r"Highlighter command .* does not exist.") begin
+                sprint() do io
+                    Cthulhu.highlight(io, "INPUT", lexer, config)
+                end == "INPUT"
+            end
+        end
+    end
+end
+
+if VERSION >= v"1.1-"
```
tkf: This is to workaround `MethodError: no method matching rawhandle(::Base.GenericIOBuffer{Array{UInt8,1}})` in Julia 1.0 https://travis-ci.org/JuliaDebug/Cthulhu.jl/jobs/556538119#L240

src/ui.jl
```diff
@@ -42,7 +42,7 @@ function TerminalMenus.header(m::CthulhuMenu)
     m.sub_menu && return ""
     """
     Select a call to descend into or ↩ to ascend. [q]uit.
-    Toggles: [o]ptimize, [w]arn, [d]ebuginfo.
+    Toggles: [o]ptimize, [w]arn, [d]ebuginfo, [s]yntax highlight for LLVM/Native.
     Show: [L]LVM IR, [N]ative code
```
tkf: I didn't use `[h]ighlight` as it may make sense to use it for `[h]elp` at some point. I guess another option would be `Show: [L]LVM IR, [N]ative code (toggle: [H]ighlight)`.

tkf:
Thanks for the review, and thanks again for this cool package!

PR #57, opened 2019-08-13
Add ability to show souce code and AST

### LINKED ISSUE #47: Display Source and AST using CodeTracking
Need to be robust against Revise not being loaded (and these thus not being defined)

oxinabox:
Close #47 

It also reorganizes some of the displaying things code.
idk if it is actually an improvement.
At some point someone should think about how to do Model+ViewController seperation in julia.
If you hate it, can always just cut this back to the minimal changes needed for just displaying source / AST

Screenshouts:

Source:
image

AST:
image

idk if that is the best way to render the AST.
See commments in code.

TODO: Tests

@TimHoly may find this of interest.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -109,14 +109,8 @@ const descend = descend_code_typed
 # src/reflection.jl has the tools to discover methods
 # src/ui.jl provides the user facing interface to which _descend responds
 ##
-function _descend(mi::MethodInstance; iswarn::Bool, params=current_params(), optimize::Bool=true, kwargs...)
+function _descend(mi::MethodInstance; iswarn::Bool, params=current_params(), optimize::Bool=true, debuginfo=true, kwargs...)
     display_CI = true
-    debuginfo = true
-    if :debuginfo in keys(kwargs)
-        selected = kwargs[:debuginfo]
-        # TODO: respect default
-        debuginfo = selected == :source
-    end
 
     while true
         (CI, rt, slottypes) = do_typeinf_slottypes(mi, optimize, params)
@@ -162,8 +156,9 @@ function _descend(mi::MethodInstance; iswarn::Bool, params=current_params(), opt
             if next_mi === nothing
                 continue
             end
+             
             _descend(next_mi; params=params, optimize=optimize,
-                     iswarn=iswarn, debuginfo=debuginfo_key, kwargs...)
+                     iswarn=iswarn, debuginfo=debuginfo, kwargs...)
 
         elseif toggle === :warn
             iswarn ⊻= true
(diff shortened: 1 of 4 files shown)


vchuravy:
Thanks I think this is a rather good improvement overall. The ast printing is to verbose though.

I will have to deal with the inevitable merge-conflict in #46, but that is my problem for not getting that merged sooner.

src/Cthulhu.jl
```diff
@@ -108,34 +109,15 @@ const descend = descend_code_typed
 # src/reflection.jl has the tools to discover methods
 # src/ui.jl provides the user facing interface to which _descend responds
 ##
-function _descend(mi::MethodInstance; iswarn::Bool, params=current_params(), optimize::Bool=true, kwargs...)
+function _descend(mi::MethodInstance; iswarn::Bool, params=current_params(), optimize::Bool=true, debuginfo=true, kwargs...)
```
vchuravy: `debuginfo` is not a boolean, it is a symbol that is either `:source`, `:default`, or `:none` (I think :P). Potentially there are other extensions like `:compact` in the future

oxinabox: I am going to need to look at this again.
Sometimes it changes?
Maybe I should pull those changes out.

I would like to see a bunch of the UI code seperated.

oxinabox:
> The ast printing is to verbose though.

That is how AST printing is though.
Sometimes that's useful.

Could do `sexpr` (as commented on in code). Which is more compact but also not a thing most people use.
Idk.

vchuravy:
> That is how AST printing is though.
> Sometimes that's useful.

I prefer the pretty-printing, but I am fine with having a verbose mode.
E.g. I prefer `show` over `dump` for quick consumption.

oxinabox:
`show` is basically the same as showing the source code though isn't it?
And we have that option.

vchuravy:
> show is basically the same as showing the source code though isn't it?

Not if the source code isn't available because it was generated.

oxinabox:
> Not if the source code isn't available because it was generated.

I think in those cases CodeTracking already falls back to doing this.

vchuravy:
Not that I can see https://github.com/timholy/CodeTracking.jl/blob/3df13a75f77ca1a2af38e5e9cea72afc4edd4f83/src/CodeTracking.jl#L191 and https://github.com/timholy/CodeTracking.jl/blob/3df13a75f77ca1a2af38e5e9cea72afc4edd4f83/src/utils.jl#L60

vchuravy:
Thank you Lyndon!

oxinabox:
Cool, well I am sure you will find out what I broken soon.
(I think with some recent changes to CodeTracking, nothing and it just works)

PR #60, opened 2019-08-26
Decode cufunction with kwargs

vchuravy:
Fixes https://github.com/vchuravy/GPUifyLoops.jl/issues/95

@mwarusz could you give this a spin?

### DIFF
--- a/src/codeview.jl
+++ b/src/codeview.jl
@@ -39,10 +39,13 @@ function cthulhu_ast(io::IO, mi, optimize, debuginfo, params, config::CthulhuCon
     meth = mi.def
     ast = definition(Expr, meth)
     if ast!==nothing
-        dump(io, ast; maxdepth=typemax(Int))
-        # Or should use: ?
-        #Meta.show_expr(io, ast)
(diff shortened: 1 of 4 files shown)


src/codeview.jl
```diff
@@ -39,10 +39,13 @@ function cthulhu_ast(io::IO, mi, optimize, debuginfo, params, config::CthulhuCon
     meth = mi.def
     ast = definition(Expr, meth)
     if ast!==nothing
-        dump(io, ast; maxdepth=typemax(Int))
-        # Or should use: ?
-        #Meta.show_expr(io, ast)
-        # Could even highlight the above as some kind-of LISP
+        if config.pretty_ast
```
oxinabox: I believe this should be `if !config.pretty_ast`?

I still think the whole way the UI works needs to be rethought.
It is all very adhoc

vchuravy: :shrug: I have yet to encounter a technology that is not adhoc. Give me a sensible TUI framework in Julia and I will rewrite the UI.
I care much more about the functionality.

PR #67, opened 2019-10-04
Small improvements and fix descend into `Union{f, g}`

vchuravy:


src/Cthulhu.jl
```diff
@@ -169,6 +169,10 @@ function _descend(mi::MethodInstance; iswarn::Bool, params=current_params(), opt
 
             if callsite.info isa MultiCallInfo
                 sub_callsites = map(ci->Callsite(callsite.id, ci), callsite.info.callinfos)
+                if isempty(sub_callsites)
+                    @warn "Found MultiCallInfo, but sub callinfos is empty"
```
oxinabox: this tells me *nothing* as a user.
Should it also tell some other things like "plz make bug report?"

vchuravy: IMO "plz make a bug report" is always implied when you receive a warning from code that you didn't write...

PR #70, opened 2020-03-07
Fix: _dump_function_linfo not defined

tkf:
`InteractiveUtils._dump_function_linfo` is gone as of https://github.com/JuliaLang/julia/pull/25984. We need to use new functions `_dump_function_linfo_llvm` and `_dump_function_linfo_native`.


vchuravy:
Thanks!

PR #76, opened 2020-04-25
Two fixes for find_callsites

tkf:
I'll try to come up with MWEs if needed before merging this.

### DIFF
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -61,7 +61,7 @@ function find_callsites(CI, mi, slottypes; params=current_params(), kwargs...)
                 callinfo = FailedCallInfo(sig, rt)
             end
         else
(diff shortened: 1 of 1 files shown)


src/reflection.jl
```diff
@@ -58,7 +61,7 @@ function find_callsites(CI, mi, slottypes; params=current_params(), kwargs...)
                 callinfo = FailedCallInfo(sig, rt)
             end
         else
-            callinfo = FailedCallInfo((ft, argTs), rt)
+            callinfo = FailedCallInfo(Tuple{ft, argTs}, rt)
```
vchuravy, line 64:
```suggestion
            callinfo = FailedCallInfo(Base.signature_type(ft, argTs), rt)
```

tkf:
Thanks for the review. I'll try to find an MWE.

PR #79, opened 2020-05-16
Add support for AbstractInterpreter

timholy:
*Maybe* I could have done a little more code-sharing, but over time I suspect these two may diverge more anyway so it seems to make sense to implement `do_typeinf_slottypes` in two versions.


vchuravy:
Thanks Tim!

PR #80, opened 2020-06-03
Show variables in warntype (fixes #40)

### LINKED ISSUE #40: Display slot types
When printing code chunks such as

```
Body::Any
1 ─       Core.NewvarNode(:(#1070))
│   %2  = Zygote.size(x)::Tuple{Int64,Int64}
│   %3  = Zygote.size(x̄)::Any
│   %4  = (%2 == %3)::Any
└──       goto #3 if not %4
2 ─       return x̄
```

It would be very helpful to be able to display the type of `x̄`, so as to be able to see the difference between the inferred types of `%2` and `%3`.

Grand Master Valentin says:

> `@code_warntype` can do that, so we could probably add that printing

timholy:


timholy:
Any objections?

PR #85, opened 2020-08-12
Small usability enhancements

timholy:
This makes small tweaks to the user experience. The only one that needs explanation is that sometimes `ascend` has a hard time finding a call to the parent due to optimization, so this makes sure we check both optimized and unoptimized instances.


vchuravy:
Thanks Tim!

PR #86, opened 2020-08-21
Improve inferrability of several methods

timholy:
Because most invalidations come from poor inference, Cthulhu is now a crucial part of the apparatus for finding and fixing invalidations. Recently I released MethodAnalysis 0.3 which allows one to find all callers of a particular MethodInstance. Interestingly, many problems often lead back to Cthulhu itself. Consequently it's important for the internals of this package to start ensuring better inference.


timholy:
`miinner` is to avoid Julia #15276; the parser gets confused having an inner function with a local variable of the same name as one in the outer scope.

PR #89, opened 2020-09-06
Add option to omit stable code when using descend_warn_type

YingboMa:
We have
```julia
julia> function foo()
           T = rand() > 0.5 ? Int64 : Float64
           sum(rand(T, 100))
       end

julia> @descend_code_warntype verbose=false foo()

│ ─ %-1  = invoke foo()::Union{Float64, Int64}
Variables
  #self#::Core.Compiler.Const(foo, false)
  T::Union{Type{Float64}, Type{Int64}}
  @_3::Union{Type{Float64}, Type{Int64}}

Body::Union{Float64, Int64}
│    @ REPL[6]:2 within `foo'
│    %35  = φ (#10 => %32, #11 => %34)::Union{Type{Float64}, Type{Int64}}
│    @ REPL[6]:3 within `foo'
│    %36  = Main.rand(%35, 100)::Union{Array{Float64,1}, Array{Int64,1}}
│    %129 = φ (#30 => %73, #49 => %118)::Union{Float64, Int64}

Toggles: [o]ptimize, [w]arn, [v]bose printing for warntype code, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.

 • %4  = invoke default_rng(::Int64)::Random.MersenneTwister
   %15  = invoke dsfmt_fill_array_close1_open2!(::Random.DSFMT.DSFMT_state,::Ptr{Float64},::Int64)::Any
   %36  = invoke rand(::Union{Type{Float64}, Type{Int64}},::Int64)::Union{Array{Float64,1}, Array{Int64,1}}
   %71  = invoke mapreduce_impl(::typeof(identity),::typeof(Base.add_sum),::Array{Float64,1},::Int64,::Int64,::Int64)
   %116  = invoke mapreduce_impl(::typeof(identity),::typeof(Base.add_sum),::Array{Int64,1},::Int64,::Int64,::Int64)
   ↩
```

### DIFF
--- a/src/ui.jl
+++ b/src/ui.jl
@@ -48,7 +48,7 @@ function TerminalMenus.header(m::CthulhuMenu)
     m.sub_menu && return ""
     """
     Select a call to descend into or ↩ to ascend. [q]uit. [b]ookmark.
-    Toggles: [o]ptimize, [w]arn, [v]bose printing for warntype code, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.
+    Toggles: [o]ptimize, [w]arn, [v]erbose printing for warntype code, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.
     Show: [S]ource code, [A]ST, [L]LVM IR, [N]ative code
     Actions: [E]dit source code, [R]evise and redisplay
     Advanced: dump [P]arams cache.
--- a/src/ui.jl
+++ b/src/ui.jl
@@ -48,7 +48,7 @@ function TerminalMenus.header(m::CthulhuMenu)
     m.sub_menu && return ""
     """
     Select a call to descend into or ↩ to ascend. [q]uit. [b]ookmark.
-    Toggles: [o]ptimize, [w]arn, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.
+    Toggles: [o]ptimize, [w]arn, [v]bose printing for warntype code, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.
     Show: [S]ource code, [A]ST, [L]LVM IR, [N]ative code
     Actions: [E]dit source code, [R]evise and redisplay
     Advanced: dump [P]arams cache.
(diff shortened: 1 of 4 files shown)


src/ui.jl
```diff
@@ -48,7 +48,7 @@ function TerminalMenus.header(m::CthulhuMenu)
     m.sub_menu && return ""
     """
     Select a call to descend into or ↩ to ascend. [q]uit. [b]ookmark.
-    Toggles: [o]ptimize, [w]arn, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.
+    Toggles: [o]ptimize, [w]arn, [v]bose printing for warntype code, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.
```
vchuravy, line 51:
```suggestion
    Toggles: [o]ptimize, [w]arn, [v]erbose printing for warntype code, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.
```

vchuravy:
Can you also add a docstring and maybe a note in the REPL? Took me a second to understand what you are doing.

Also since you are adding quite a bit of new code, it would be good if you could unit-tests that.

vchuravy (approved):

timholy:
I love this! I think we should probably merge this as-is, but out of curiosity, did you consider using `show_ir_stmt` and restricting this functionality to Julia 1.6 and higher? I'm just a little worried about the maintenance of this code. Of course, you can say that the less we rely on Base/Core, perhaps the more robust we become, but conversely any changes to the IR require us to update code here.

Bottom line, I don't even know what the right path is. I'd be curious to hear your thoughts, but no matter what, nothing needs to stand in the way of merging this.

YingboMa:
I did notice the `show_ir_stmt` function in Julia 1.6, but for the sake of backward compatibility, I chose to use the old structure.

timholy:
Seems fine. If we start having maintenance issues with the current implementation we can revisit that idea. Really, once 1.6 becomes the LTS I plan to entirely drop support in my packages for any lesser version :wink: .

PR #90, opened 2020-10-04
Fix functionality and tests on nightly

timholy:
The changes to the tests might be https://github.com/JuliaLang/julia/pull/37714? Not sure.

@vchuravy, I was worried that this avoids testing some functionality that would have been tested before, but curiously with code-coverage I get no differences in `callsite.jl` and only very minor differences between 1.6 and 1.5 in total coverage (207 out of 676 eligible lines vs 204 out of 677 lines, respectively).


vchuravy:
The testing strategy for Cthulhu is quite haphazard in the first place... and dependent on internals of Julia.

PR #91, opened 2020-10-05
Support arrays in splatting

timholy:
Demo:
```julia
julia> function fsplat(::Type{Int}, a...)
           z = zero(Int)
           for v in a
               z += v
           end
           return z
       end
fsplat (generic function with 1 method)

julia> gsplat1(T::Type, a...) = fsplat(T, a...)   # does not force specialization
gsplat1 (generic function with 1 method)

julia> hsplat1(A) = gsplat1(eltype(A), A...)
hsplat1 (generic function with 1 method)

julia> using Cthulhu

julia> @descend hsplat1([1,2,3])

│ ─ %-1  = invoke hsplat1(::Vector{Int64})::Int64
CodeInfo(
    @ REPL[3]:1 within `hsplat1'
1 ─ %1 = Core.tuple(Int64)::Core.Const((Int64,))
│   %2 = Core._apply_iterate(Base.iterate, Main.gsplat1, %1, A)::Int64
└──      return %2
)
Toggles: [o]ptimize, [w]arn, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.

 • %2  = invoke gsplat1(::DataType,::Vararg{Int64, N} where N)::Int64
```

Without this PR you can't descend into `gsplat1`.

Note this isn't quite perfect: the first argument's type is `DataType` despite the `Core.Const((Int64,))` annotation of `%1`, but that will be the subject of a later PR. The other aspect that is not perfect, but which I'm not sure how to solve, is that doing something like `foo([1,2,3]..., 4)` is perfectly legal julia code, it's just that julia won't know how long the vararg is. I guess the best answer might be to `typejoin` or `Union` the type in `Vararg`, but that's not implemented here.

EDIT: only works on Julia 1.4+. Presumably one could do something similar for earlier Julia versions, but meh.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,7 +18,8 @@ julia = "1"
 
 [extras]
 Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "Revise"]
+test = ["Test", "Revise", "StaticArrays"]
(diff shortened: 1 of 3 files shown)


vchuravy:
>  Presumably one could do something similar for earlier Julia versions, but meh.

Agreed ;)

PR #92, opened 2020-10-09
Improve treatment of types-as-arguments & specialization heuristics

timholy:
On current master, using test functions https://github.com/JuliaDebug/Cthulhu.jl/blob/4c28b946d6e1c434d66c951c1c1cdd3920a71bf6/test/runtests.jl#L142-L150

one gets
```julia
julia> @descend hsplat1([1,2,3])

│ ─ %-1  = invoke hsplat1(::Vector{Int64})::Int64
CodeInfo(
    @ REPL[3]:1 within `hsplat1'
1 ─ %1 = Core.tuple(Int64)::Core.Const((Int64,))
│   %2 = Core._apply_iterate(Base.iterate, Main.gsplat1, %1, A)::Int64
└──      return %2
)
Toggles: [o]ptimize, [w]arn, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.

 • %2  = invoke gsplat1(::DataType,::Vararg{Int64, N} where N)::Int64
```
There's something a bit weird here: in the `call` section, in the `gsplat1` call, the first argument is annotated `::DataType` even though at %1 it's `Const((Int64,))`. This does not reflect some deep understanding that `gsplat1` does not specialze on its first argument, because the same holds for
```julia
gsplat2(::Type{T}, a...) where T = fsplat(T, a...)
hsplat2(A) = gsplat2(eltype(A), A...)
```
which does force specialization.

This PR changes the result to
```julia
julia> @descend hsplat1([1,2,3])

│ ─ %-1  = invoke hsplat1(::Vector{Int64})::Int64
CodeInfo(
    @ REPL[3]:1 within `hsplat1'
1 ─ %1 = Core.tuple(Int64)::Core.Const((Int64,))
│   %2 = Core._apply_iterate(Base.iterate, Main.gsplat1, %1, A)::Int64
└──      return %2
)
Toggles: [o]ptimize, [w]arn, [d]ebuginfo, [s]yntax highlight for Source/LLVM/Native.

 • %2  = invoke gsplat1(::Type{Int64},::Vararg{Int64, N} where N)::Int64
   ↩
```

The reason I've marked this WIP is that this version *also* doesn't distinguish between `hsplat1` and `hsplat2`: instead of `::DataType` for both it's `::Type{Int}` for both. Yet they clearly differ:
```julia
julia> using BenchmarkTools

julia> @btime hsplat1([1,2,3])
  847.803 ns (6 allocations: 336 bytes)
6

julia> @btime hsplat2([1,2,3])
  314.957 ns (4 allocations: 176 bytes)
6
```
I confess I'm not sure how this should be fixed. Any ideas?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,9 +17,10 @@ FoldingTrees = "1"
 julia = "1"
 
 [extras]
+MethodAnalysis = "85b6ec6f-f7df-4429-9514-a64bcd9ee824"
 Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "Revise", "StaticArrays"]
+test = ["Test", "MethodAnalysis", "Revise", "StaticArrays"]
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -90,7 +90,7 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
                 end
             elseif c.head === :call
                 rt = CI.ssavaluetypes[id]
-                types = mapany(arg -> widenconst(argextype(arg, CI, sptypes, slottypes)), c.args)
+                types = mapany(arg -> widenconst_t(argextype(arg, CI, sptypes, slottypes)), c.args)
 
                 # Look through _apply
                 ok = true
@@ -194,6 +194,19 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
     end
     return callsites
 end
+
+function widenconst_t(c::Const)
+    if isa(c.val, Type)
+        if Core.Compiler.isvarargtype(c.val)
+            return Type
+        end
+        return Type{c.val}
+    else
+        cv = c.val
+        return isa(cv, Tuple) ? Tuple{map(Core.Typeof, cv)...} : Core.Typeof(cv)
+    end
+end
+widenconst_t(t) = widenconst(t)
 
 if VERSION >= v"1.1.0-DEV.215"
 function dce!(ci, mi)
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -302,14 +302,23 @@ function choose_method_instance(@nospecialize(F), @nospecialize(TT); params=curr
     choose_method_instance(sig; params=params)
 end
 
-function callinfo(sig, rt, max=-1; params=current_params())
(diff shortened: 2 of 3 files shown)


vchuravy:
Yeah the challenge is to differentiating between callsite knowledge, and callee constraints. So if we want to differentiate between the two we need to look at the callee and probably look at the `spvals`.

timholy:
Where I'm stuck is that method lookup doesn't seem to reveal the issue:
```julia
julia> sig = Tuple{typeof(gsplat1), Type{Int64}, Vararg{Int64, N} where N}
Tuple{typeof(gsplat1), Type{Int64}, Vararg{Int64, N} where N}

julia> methds = Base._methods_by_ftype(sig, -1, typemax(UInt))
1-element Vector{Any}:
 Core.MethodMatch(Tuple{typeof(gsplat1), Type{Int64}, Vararg{Int64, N} where N}, svec(), gsplat1(T::Type, a...) in Main at REPL[3]:1, true)

julia> mm = methds[1]
Core.MethodMatch(Tuple{typeof(gsplat1), Type{Int64}, Vararg{Int64, N} where N}, svec(), gsplat1(T::Type, a...) in Main at REPL[3]:1, true)

julia> mm.spec_types
Tuple{typeof(gsplat1), Type{Int64}, Vararg{Int64, N} where N}

julia> mm.sparams
svec()


julia> mm.method
gsplat1(T::Type, a...) in Main at REPL[3]:1

julia> mm.fully_covers
true
```

Yet if you call `hsplat1` and `hsplat2` you get this:
```julia
julia> using MethodAnalysis

julia> methodinstances(gsplat1)
 MethodInstance for gsplat1(::Type, ::Int64, ::Vararg{Int64, N} where N)
 MethodInstance for gsplat1(::Type{Int64}, ::Vararg{Int64, N} where N)

julia> methodinstances(gsplat2)
 MethodInstance for gsplat2(::Type{Int64}, ::Int64, ::Vararg{Int64, N} where N)
 MethodInstance for gsplat2(::Type{Int64}, ::Vararg{Int64, N} where N)
```

To me it seems difficult to fix this unless `_methods_by_ftype` were to start returning an unspecialized signature.

timholy:
OK, this one seems to do the trick (on 1.6). The one thing I worry a bit about is the *lack* of specialization in things like `hsplat1((1,2,3))` when optimization is off (which is why I had to change the `hsplat1` test result)...ideally it would reflect the actual call chain but avoid inlining.

timholy:
If we merge this, it should be released as 1.3, as the behavior is pretty significantly changed---hopefully, improved---for some calls.

timholy:
This is being superseded in a much better way by @Keno's work.

PR #96, opened 2020-10-28
Add dump_module option to code_llvm(::IO, ::Bookmark)

tkf:
This PR adds keyword option `dump_module` to `cthulhu_llvm(...)` and `code_llvm(io, bookmark)` so that the latter behaves like stdlib's `code_llvm(f, sig)`.

With this PR and https://github.com/tkf/CodeX.jl/compare/cthulhu, we can do, for example,

```julia
c = CodeX.llvm(Cthulhu.BOOKMARKS[END])
c.cfg_only         # view CFG
c.godbolt()        # open LLVM in godbolt
c.native.godbolt() # open ASM in godbolt
```

### DIFF
--- a/src/codeview.jl
+++ b/src/codeview.jl
@@ -15,18 +15,19 @@ function highlight(io, x, lexer, config::CthulhuConfig)
     end
 end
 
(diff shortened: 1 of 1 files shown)


vchuravy:
Thanks!

PR #97, opened 2020-10-29
Fix jl_new_task handling

tkf:
...and improve the test by making sure that `TaskCallInfo` does not wrap a `FailedCallInfo`.

### DIFF
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -180,9 +180,10 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
                 if c.args[1] isa QuoteNode
                     cfunc = c.args[1].value
                     if cfunc === :jl_new_task
-                        func = c.args[7]
+                        func = c.args[6]
                         ftype = widenconst(argextype(func, CI, sptypes, slottypes))
-                        callsite = Callsite(id, TaskCallInfo(callinfo(ftype, nothing, params=params)))
+                        sig = Tuple{ftype}
+                        callsite = Callsite(id, TaskCallInfo(callinfo(sig, nothing, params=params)))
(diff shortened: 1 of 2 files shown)


src/reflection.jl
```diff
@@ -180,9 +180,10 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
                 if c.args[1] isa QuoteNode
                     cfunc = c.args[1].value
                     if cfunc === :jl_new_task
-                        func = c.args[7]
+                        func = c.args[6]
```
vchuravy, line 183: Can we add a version check? Since this used to work.

tkf: Hmm... Do you remember with which version it worked? [Julia 1.0.5 passes the test on Travis.](https://travis-ci.org/github/JuliaDebug/Cthulhu.jl/jobs/739815528#L175)

Also, I checked it with 1.0.5 in REPL:

```julia
julia> VERSION
v"1.0.5"

(v1.0) pkg> activate .

shell> git log -n3 --oneline
c3b884b (HEAD -> master) Merge branch 'deoptimized' into wd/amdci2
39e7b57 Support descending into deoptimized calls
5b20a73 Fix jl_new_task handling

julia> g() = schedule(Task(h))
       h() = nothing
h (generic function with 1 method)

julia> Base.print(io::IO, ::Nothing) = print(io, "nothing")

julia> using Cthulhu

julia> @descend g()

...

 • %2  = task < h()::nothing >
   %14  = deoptimized #throw_inexacterror(::QuoteNode,::Type{Int64},::Int64)::Union{}
   %32  = deoptimized error(::String)::Union{}
   ↩
```

PR #98, opened 2020-10-29
Support descending into deoptimized and non-deoptimized calls

tkf:
Update: In response to @vchuravy's comment https://github.com/JuliaDebug/Cthulhu.jl/pull/98#issuecomment-718813884, this PR now lets the user decide the method to descend into. So, it's possible to look at the method that would be available in the current released version.

---

Sometimes, `julia` decides ("deoptimize") not use the full type information available. As a result, the `MethodInstance` in the `:invoke` instruction has less accurate signature than the one available through `CI.ssavaluetypes`. I suggest to use more accurate signature whenever possible during descend.

When combining this PR with #97, we can descend into the loop body of `Threads.@threads`. For example, with

```julia
f() = Threads.@threads for i in 1:10; noinline(); end
@noinline noinline() = read("/dev/null")
```

`@descend f()` now shows

```
 • %23  = deoptimized getindex(::Tuple{},::Int64)::Union{}
   %34  = deoptimized threading_run(::var"#1856#threadsfor_fun#28"{UnitRange{Int64}})::Any
   ↩
```

while before it was

```
 • %23  = invoke getindex(::Tuple,::Int64)::Union{}
   %34  = invoke threading_run(::Function)::Any
   ↩
```

### DIFF
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -29,6 +29,15 @@ else
     code_for_method(method, metharg, methsp, world, force=false) = Core.Compiler.specialize_method(method, metharg, methsp, force)
 end
 
+# https://github.com/JuliaLang/julia/pull/36318
+_id(x) = x.id
+if VERSION < v"1.6.0-DEV.272"
+    const SlotOrArgument = Core.Slot
+else
+    _id(x::Core.Argument) = x.n
+    const SlotOrArgument = Union{Core.Argument,Core.Slot}
+end
+
 transform(::Val, callsite) = callsite
 function transform(::Val{:CuFunction}, callsite, callexpr, CI, mi, slottypes; params=nothing, kwargs...)
     sptypes = sptypes_from_meth_instance(mi)
@@ -91,9 +100,12 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
                     argtypes_ssa = map(c.args[3:end]) do a
                         if isa(a, Core.SSAValue)
                             a = CI.ssavaluetypes[a.id]
-                            if isa(a, Const)
-                                a = a.val
-                            end
+                            isa(a, Const) || return a
+                            a = a.val
+                        elseif isa(a, SlotOrArgument)
+                            a = CI.slottypes[_id(a)]
+                            isa(a, Const) || return a
+                            a = a.val
                         end
                         Core.Typeof(a)
                     end
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -103,7 +103,7 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
                             isa(a, Const) || return a
                             a = a.val
                         elseif isa(a, SlotOrArgument)
-                            a = CI.slottypes[_id(a)]
+                            a = slottypes[_id(a)]
                             isa(a, Const) || return a
                             a = a.val
                         end
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -111,7 +111,10 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
                     end
                     sig_ssa = Tuple{Base.tuple_type_head(mi.def.sig), argtypes_ssa...}
                     if sig_ssa !== mi.def.sig
-                        callsite = Callsite(id, DeoptimizedCallInfo(callinfo(sig_ssa, rt), callsite.info))
+                        sig_callinfo = callinfo(sig_ssa, rt)
+                        if get_mi(sig_callinfo) !== mi
+                            callsite = Callsite(id, DeoptimizedCallInfo(sig_callinfo, callsite.info))
+                        end
                     end
                 end
             elseif c.head === :call
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -87,6 +87,20 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
                 mi = get_mi(callsite)
                 if nameof(mi.def.module) === :CUDAnative && mi.def.name === :cufunction
                     callsite = transform(Val(:CuFunction), callsite, c, CI, mi, slottypes; params=params, kwargs...)
+                elseif callsite.info isa MICallInfo
+                    argtypes_ssa = map(c.args[3:end]) do a
+                        if isa(a, Core.SSAValue)
+                            a = CI.ssavaluetypes[a.id]
+                            if isa(a, Const)
+                                a = a.val
+                            end
+                        end
+                        Core.Typeof(a)
+                    end
+                    sig_ssa = Tuple{Base.tuple_type_head(mi.def.sig), argtypes_ssa...}
+                    if sig_ssa !== mi.def.sig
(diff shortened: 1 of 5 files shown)


vchuravy:
cc: @maleadt who was looking at this when working on AbstractInterpreter.

vchuravy:
IIUC Julia will actually call the de-optimized MI and not the more precise one that you want to descend into. We should offer the user the choice (similar to MultiCallInfo) which one she wants to explore. Since Julia will call the less inferred option if I were to debug an inference problem I would want to look at the actual one.

tkf:
OK, I added the submenu support:

```
   %23  = deoptimized getindex(::Tuple{},::Int64)::Union{}
 • %34  = deoptimized threading_run(::var"#30#threadsfor_fun#1"{UnitRange{Int64}})::Any
   ↩
 • %34  = invoke threading_run(::var"#30#threadsfor_fun#1"{UnitRange{Int64}})::Any
   %34  = invoke threading_run(::Function)::Any
   ↩
```

timholy:
Aside from details about type arguments, it's interesting this seems to go in almost the exact opposite direction of #92, which aims (but does not yet entirely succeed) to reduce "spurious" specialization available only in the caller. I can see a role for this, though, and the menu is a great idea if we go with this. But can you amplify

>  I suggest to use more accurate signature whenever possible during descend.

Why is that better?

tkf:
Hi Tim, thanks for the review!

I probably should've updated the OP. As I noted in my previous comment, it now gives the user a choice to descend into different methods. So, it preserves an option to descend into the one that you can do in the current released version.

> it's interesting this seems to go in almost the exact opposite direction of #92

So, I don't think the current version is the opposite of #92. It's a union of the current behavior and the behavior I need. If you want to add a new descent "branch" or replace the current deoptimized direction of descent, it's very easy to do with the updated PR now.

> > I suggest to use more accurate signature whenever possible during descend.
> 
> Why is that better?

As I noted in the OP, one of the motivations is to descend into the accurate loop body of `Threads.@threads`. I don't think we can look at the loop body with current Cthulhu.

vchuravy:
> As I noted in the OP, one of the motivations is to descend into the accurate loop body of Threads.@threads. I don't think we can look at the loop body with current Cthulhu.

Actually we can, but only by happenstance. There is a call to the functionbody in a different branch,

tkf:
By "current Cthulhu", I meant the latest released version or at least the main branch. I'd love to have a solution that does not involve maintaining a fork myself.

I also find that it's pretty informative to see where Julia "gave up." Since this PR only adds information (of decent quality; not noises) and preserve the ones that are already available, what's the downside?

vchuravy:
Sorry I should have been clearer, it used to be the case that there was a direct call to the threadsfun that wasn't de-optimized in fast-path branch. That code has changed since then and that is no longer true.

src/reflection.jl
```diff
@@ -87,6 +87,20 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
                 mi = get_mi(callsite)
                 if nameof(mi.def.module) === :CUDAnative && mi.def.name === :cufunction
                     callsite = transform(Val(:CuFunction), callsite, c, CI, mi, slottypes; params=params, kwargs...)
+                elseif callsite.info isa MICallInfo
+                    argtypes_ssa = map(c.args[3:end]) do a
+                        if isa(a, Core.SSAValue)
```
vchuravy, line 92: I think you are missing Slot/Argument

vchuravy (approved):

timholy:
OK, having looked at this more, this does seem useful:
```julia
julia> @noinline function f(@nospecialize(t))
           s = 0
           for k in t
               s += k
           end
           return s
       end

julia> g() = f((1,2,3))
g (generic function with 1 method)

julia> g()
6

julia> using MethodAnalysis

julia> methodinstances(f)
 MethodInstance for f(::Any)
 MethodInstance for f(::Tuple{Int64, Int64, Int64})

julia> @code_llvm g()
;  @ REPL[2]:1 within `g'
define i64 @julia_g_424() {
top:
  %0 = alloca {}*, align 8
  store {}* inttoptr (i64 140005764089200 to {}*), {}** %0, align 8
  %1 = call nonnull {}* @j1_f_426({}* inttoptr (i64 140005748869712 to {}*), {}** nonnull %0, i32 1)
  %2 = bitcast {}* %1 to i64*
  %3 = load i64, i64* %2, align 8
  ret i64 %3
}
```
So this PR will give you the option to descend into the real call, and that is predictive of the MethodInstances that get generated and what LLVM knows.

Could use a test. This `f`/`g` pair could serve in that capacity?

src/reflection.jl
tkf: Thanks for catching it! I added the support for Core.Argument and Core.Slot:

https://github.com/JuliaDebug/Cthulhu.jl/blob/610abb4aa85822dde850851058314873509239e9/src/reflection.jl#L32-L39

https://github.com/JuliaDebug/Cthulhu.jl/blob/610abb4aa85822dde850851058314873509239e9/src/reflection.jl#L105-L108

tkf: Apparently, we cannot use `CI.slottypes` because of Julia < 1.2? It's now:

https://github.com/JuliaDebug/Cthulhu.jl/blob/e2ec4434b01115cf265505b11b1490b930de979d/src/reflection.jl#L105-L106

tkf:
Good to know that you think it could be useful!  I added a test based on your example (+ another one for more coverage).

Locally it now works with 1.0. I'll merge this once Travis is green.

timholy:
Just to clarify a remark I made above,, IIUC the native code that implements `f` will correspond to `f(::Any)` but the caller `g` exploits the inference result on `f(::Tuple{Int,Int,Int})` to predict the return type and make inlining decisions (and the `@nospecialize` has no effect if it does get inlined). Consequently anything "above" `f` will be best informed by `f(::Tuple{Int,Int,Int})`, and anything in `f` or "below" (one of its callees) will be best informed by `f(::Any)`. Consequently I still suspect that descending into the deoptimized version may generally be more informative for the purpose of performance analysis, but it will still be nice to be able to compare these.

timholy (approved):

vchuravy:
> IIUC the native code that implements f will correspond to f(::Any)

It the case of function types (and probably `@nospecialize`) it will actually represent the de-optimized IR. Codegen generally will place direct calls to the code represented by the MethodInstance and only in some cases fall back to a generic call, which would then use the callsite information.

vchuravy:
Thanks @tkf!

timholy:
I'd like to make sure I understand https://github.com/JuliaDebug/Cthulhu.jl/pull/98#issuecomment-719975299, but there seem to be one or two typos that make it a little hard to parse. Can I ask for an edit?

vchuravy:
I should not be allowed on the internet without Coffee. I had trouble re-reading my statement and I think I misunderstood your statement. 

Julia's Codegen will use the specific method instance when emitting an `invoke`, therefore the executed code will correspond the the `f(::Any)` or `f(::Function)`. 

There is a fallback in the code https://github.com/JuliaLang/julia/blob/ed5fd937c958449a88054e6ee2b90460ddb52bab/src/codegen.cpp#L3467 that would lead to `jlcall` which I miss-remembered as being a generic call.

PR #108, opened 2020-12-20
Handle PartialStruct

YingboMa:
I am not sure how to create a test for this.

### DIFF
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -38,6 +38,24 @@ else
     const SlotOrArgument = Union{Core.Argument,Core.Slot}
 end
 
+if VERSION < v"1.2.0-DEV.354"
+    function unwrapconst(a)
+        if isa(a, Const)
+            a = Core.Typeof(a.val)
+        end
+        return a
+    end
+else
+    function unwrapconst(a)
+        if isa(a, Const)
+            a = Core.Typeof(a.val)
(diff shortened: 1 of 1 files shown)


timholy:
Very nice. For a test, can you design a function that produces a `PartialStruct` and then check for it using `find_callsites_by_ftt`?

YingboMa:
I don't know how to write a small function that produces a `PartialStruct `. I discovered this when I was working on a pretty large codebase. Do you mind to write an example? I'd be happy to include it in this PR.

vchuravy (approved):

timholy:
I think the best bet would be to hold off on merging this and locally have it throw `error("I got a PartialStruct, WHOOPPEEE!")` when it gets to PartialStruct. That way we'll notice when it happens and then we can add a test. Surely there's something small out there that generates one, and we'll just come across it in the ordinary course of business.

YingboMa:
Since this fixes my use case, do you mind to merge it for now and make a release?

timholy:
OK, just be sure to submit a test case if you stumble across something workable as a test.

timholy:
I'd love to include #109, but I will leave it open for a bit longer and see if anyone wants to review it.

timholy:
https://github.com/JuliaRegistries/General/pull/26833

PR #109, opened 2020-12-22
Add `ascend(bt)` where bt is a backtrace

timholy:


timholy:
I also added some documentation on `ascend`.

vchuravy (approved):

PR #111, opened 2020-12-23
Add link to JuliaCon video

oxinabox:
Linking to the video is useful for people that like that.
I put it at the top, but might be better to put it at the bottom. idk.


vchuravy:
Thanks!

timholy:
I'm really glad you did this. When I went looking for Val's video I seem to have found something else much less useful. It's a great overview and tutorial.

PR #113, opened 2020-12-27
Preserve the Expr head when printing callsites

timholy:
It's useful to be able to distinguish between compile-time and
run-time dispatch. Internally Julia does this with the `:invoke`/`:call`
distinction, but Cthulhu has historically used `:invoke` for both,
only using `:call` when multiple MethodInstances might apply.
Unfortunately, the multi-MI analysis does not take full account of
specialization, and so does not fully replicate the original pattern.

Here, we directly preserve the original head, which should make it
more robust to mistakes or changes in how Julia does things.

## Demo

On the current `master` branch,
```julia
julia> @noinline twice(x::Real) = 2x
twice (generic function with 1 method)

julia> calltwice(c) = twice(c[1])
calltwice (generic function with 1 method)

julia> c64, cabs = [1.0], AbstractFloat[1.0f0]
([1.0], AbstractFloat[1.0f0])

julia> @code_typed calltwice(c64)
CodeInfo(
1 ─ %1 = Base.arrayref(true, c, 1)::Float64
│   %2 = invoke Main.twice(%1::Float64)::Float64
└──      return %2
) => Float64

julia> @code_typed calltwice(cabs)
CodeInfo(
1 ─ %1 = Base.arrayref(true, c, 1)::AbstractFloat
│   %2 = Main.twice(%1)::Any
└──      return %2
) => Any
```
From the lack of `invoke` in `%2` in the latter, you know this one is a `:call` Expr:
```julia
julia> ans.first.code[2].head
:call
```

Now let's see how the `master` branch of Cthulhu handles this:
```julia
julia> using Cthulhu
[ Info: Precompiling Cthulhu [f68482b8-f384-11e8-15f7-abe071a5a75f]

julia> @descend calltwice(cabs)

│ ─ %-1  = invoke calltwice(::Vector{AbstractFloat})::Any
CodeInfo(
    @ REPL[2]:1 within `calltwice'
   ┌ @ array.jl:787 within `getindex'
1 ─│ %1 = Base.arrayref(true, c, 1)::AbstractFloat
│  └
│   %2 = Main.twice(%1)::Any
└──      return %2
)
 • %2  = invoke twice(::AbstractFloat)::Any
   ↩

```
It prints it with `invoke`. 

On this branch:
```julia
julia> @descend calltwice(cabs)

│ ─ %-1  = invoke calltwice(::Vector{AbstractFloat})::Any
CodeInfo(
    @ REPL[2]:1 within `calltwice'
   ┌ @ array.jl:787 within `getindex'
1 ─│ %1 = Base.arrayref(true, c, 1)::AbstractFloat
│  └
│   %2 = Main.twice(%1)::Any
└──      return %2
)
 • %2  = call twice(::AbstractFloat)::Any
   ↩
``` 

This mirrors the actual `MethodInstances`: in a fresh session,
```julia
julia> using MethodAnalysis

julia> calltwice(cabs)
2.0f0

julia> methodinstances(twice)
 MethodInstance for twice(::AbstractFloat)
 MethodInstance for twice(::Float32)
```
so in reality there *are* two `MethodInstance`s. The problem is that `Cthulhu.callinfo` is based on Methods, and there is only a single Method that applies, but Julia will generate specializations. I am guessing there must be a Core.Compiler method that will return all the relevant specializations, but off the top of my head I don't know what it is.

I've added a `@test_broken` to capture this failure to properly handle specialization.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -168,7 +168,7 @@ function _descend(mi::MethodInstance; iswarn::Bool, params=current_params(), opt
         display_CI = true
 
         TerminalMenus.config(cursor = '•', scroll = :wrap)
-        menu = CthulhuMenu(callsites)
+        menu = CthulhuMenu(callsites, optimize)
         cid = request(menu)
         toggle = menu.toggle
 
@@ -188,7 +188,7 @@ function _descend(mi::MethodInstance; iswarn::Bool, params=current_params(), opt
                     callsite.info.callinfos
                 end
                 sub_callsites = let callsite=callsite
-                    map(ci->Callsite(callsite.id, ci), callinfos, callsite.head)
+                    map(ci->Callsite(callsite.id, ci, callsite.head), callinfos)
                 end
                 if isempty(sub_callsites)
                     @warn "Expected multiple callsites, but found none. Please fill an issue with a reproducing example" callsite.info
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -194,7 +194,7 @@ function _descend(mi::MethodInstance; iswarn::Bool, params=current_params(), opt
                     @warn "Expected multiple callsites, but found none. Please fill an issue with a reproducing example" callsite.info
                     continue
                 end
-                menu = CthulhuMenu(sub_callsites, sub_menu=true)
+                menu = CthulhuMenu(sub_callsites, optimize; sub_menu=true)
                 cid = request(menu)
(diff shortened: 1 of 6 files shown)


timholy:
One disadvantage here is that it always uses `:call` on unoptimized code...I'm not sure that's an improvement.

timholy:
I've found a solution I'm pretty happy with (only print the Expr head for optimized code), so I'll merge this shortly barring objections.

vchuravy (approved):

PR #117, opened 2021-01-11
Unwrap `QuoteNode` arguments.

### LINKED ISSUE #116: `Error: MethodInstance extraction failed` when passing multiple `Val`s with symbols
```julia
using Cthulhu
const X = 3

@generated function foo(::Val{A}, x) where {A}
    t1 = Expr(:tuple, A...);
    Expr(:block, Expr(:meta,:noinline), :($t1.+ x))
end
bar(x) = foo(Val((:X,:X,:X)), x)
bar(2)
@descend bar(2)
```
I get:
```julia
julia> @descend bar(2)
┌ Error: MethodInstance extraction failed
│   ci.sig = Tuple{typeof(foo), QuoteNode, Int64}
│   ci.rt = Tuple{Int64, Int64, Int64}
└ @ Cthulhu ~/.julia/packages/Cthulhu/dG9TN/src/callsite.jl:19

│ ─ %-1  = invoke bar(::Int64)::Tuple{Int64, Int64, Int64}
CodeInfo(
    @ REPL[4]:1 within `bar'
1 ─ %1 = invoke Main.foo($(QuoteNode(Val{(:X, :X, :X)}()))::Val{(:X, :X, :X)}, _2::Int64)::Tuple{Int64, Int64, Int64}
└──      return %1
)
 • %1  = deoptimized #foo(::QuoteNode,::Int64)::Tuple{Int64, Int64, Int64}
```

Everything still seems to work, e.g. I can still descnd into `foo` -> invoke foo -> :
```julia
│ ─ %-1  = invoke foo(::Val{(:X, :X, :X)},::Int64)::Tuple{Int64, Int64, Int64}
CodeInfo(
    @ REPL[3]:1 within `foo'
   ┌ @ REPL[3]:1 within `macro expansion'
   │┌ @ broadcast.jl:882 within `materialize'
   ││┌ @ broadcast.jl:1095 within `copy'
   │││┌ @ ntuple.jl:50 within `ntuple'
   ││││┌ @ broadcast.jl:1095 within `#19'
   │││││┌ @ broadcast.jl:621 within `_broadcast_getindex'
   ││││││┌ @ broadcast.jl:648 within `_broadcast_getindex_evalf'
   │││││││┌ @ int.jl:87 within `+'
1 ─││││││││ %1 = Base.add_int(3, x)::Int64
│  │││││└└└
│  │││││┌ @ broadcast.jl:621 within `_broadcast_getindex'
│  ││││││┌ @ broadcast.jl:648 within `_broadcast_getindex_evalf'
│  │││││││┌ @ int.jl:87 within `+'
│  ││││││││ %2 = Base.add_int(3, x)::Int64
│  │││││└└└
│  │││││┌ @ broadcast.jl:621 within `_broadcast_getindex'
│  ││││││┌ @ broadcast.jl:648 within `_broadcast_getindex_evalf'
│  │││││││┌ @ int.jl:87 within `+'
│  ││││││││ %3 = Base.add_int(3, x)::Int64
│  ││││└└└└
│  ││││ %4 = Core.tuple(%1, %2, %3)::Tuple{Int64, Int64, Int64}
│  │└└└
└──│      return %4
   └
)
```
But it's not entirely clear what the implications of `deoptimized` are, for example, the function call and `foo` itself look normal, and the call vs invoke page doesn't appear to have any code associated with it, so it doesn't look like a dynamic dispatch as far as I can tell, and it benchmarks identically to a version without the `Val`:
```julia
julia> @noinline function nofoo(x)
           (3,3,3) .+ x
       end
nofoo (generic function with 1 method)

julia> baz(x) = nofoo(x)
baz (generic function with 1 method)

julia> @btime baz($(Ref(2))[])
  6.233 ns (0 allocations: 0 bytes)
(5, 5, 5)

julia> @btime bar($(Ref(2))[])
  6.234 ns (0 allocations: 0 bytes)
(5, 5, 5)
```

chriselrod:
Fixes #116 
That is, it fixes it assuming that the current behavior is incorrect and the new behavior correct.
```julia
julia> using Cthulhu

julia> @noinline function mysumvalarg(x, ::Val{N}) where {N}
           s = zero(eltype(x)); N
           @inbounds @simd for i ∈ eachindex(x)
               s += x[i]
           end
           s
       end
mysumvalarg (generic function with 1 method)

julia> mysumval(x) = mysumvalarg(x, Val(4))
mysumval (generic function with 1 method)

julia> x = rand(10);

julia> @descend mysumval(x)

│ ─ %-1  = invoke mysumval(::Vector{Float64})::Float64
CodeInfo(
    @ REPL[3]:1 within `mysumval'
1 ─ %1 = invoke Main.mysumvalarg(_2::Vector{Float64}, $(QuoteNode(Val{4}()))::Val{4})::Float64
└──      return %1
)
 • %1  = invoke mysumvalarg(::Vector{Float64},::Val{4})::Float64
   ↩
```

### DIFF
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -50,7 +50,7 @@ else
         if isa(a, Const)
             a = Core.Typeof(a.val)
         elseif isa(a, Core.Compiler.PartialStruct)
-            a = Core.Typeof(a.typ)
+            a = a.typ
         end
         return a
     end
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -122,6 +122,8 @@ function find_callsites(CI::Core.CodeInfo, mi::Core.MethodInstance, slottypes; p
                         elseif isa(a, SlotOrArgument)
                             a = slottypes[_id(a)]
                             return unwrapconst(a)
+                        elseif isa(a, QuoteNode)
+                            return Core.Typeof(a.value)
                         end
                         Core.Typeof(a)
                     end


chriselrod:
This doesn't fixes the `QuoteNode`s of my motivating example, but not the `PartialStruct`, because in [unwrapconst](https://github.com/JuliaDebug/Cthulhu.jl/blob/028eb04e8b85e17164b5c0343686405331e75123/src/reflection.jl#L52):
```julia
        elseif isa(a, Core.Compiler.PartialStruct)
            a = Core.Typeof(a.typ)
        end
```
`a.typ` is the correct answer/makes it work, while `Core.Typeof(a.typ)` is incorrect.

I don't know enough about the history of Julia internals nor their behavior in a particular snapshot to say what sort of special casing is needed here to get it to do `a.typ` in my case without breaking anything.

chriselrod:
For reference:
```julia
using LoopVectorization, Cthulhu
function AmulB!(C,A,B)
    @avx for n in axes(C,2), m in axes(C,1)
        Cmn = zero(eltype(C))
        for k in axes(B,1)
           Cmn += A[m,k] * B[k,n]
        end
        C[m,n] = Cmn
    end
end
M = K = N = 72;
A = rand(M,K); B = rand(K,N); C = Matrix{Float64}(undef, M, N);
ci = @code_typed AmulB!(C, A, B);
CI = ci.first;
id = 74;
c = CI.code[id];
argtypes_ssa = map(c.args[3:end]) do a
    if isa(a, Core.SSAValue)
        a = CI.ssavaluetypes[a.id]
        return Cthulhu.unwrapconst(a)
    elseif isa(a, Cthulhu.SlotOrArgument)
        a = slottypes[Cthulhu._id(a)]
        return Cthulhu.unwrapconst(a)
    elseif isa(a, QuoteNode)
        return Core.Typeof(a.value)
    end
    Core.Typeof(a)
end
acorrect = CI.ssavaluetypes[c.args[end].id].typ
code_typed(LoopVectorization._avx_!, (argtypes_ssa[1], argtypes_ssa[2], argtypes_ssa[3], argtypes_ssa[4],argtypes_ssa[5],argtypes_ssa[6]))
code_typed(LoopVectorization._avx_!, (argtypes_ssa[1], argtypes_ssa[2], argtypes_ssa[3], argtypes_ssa[4],argtypes_ssa[5], acorrect))
```
This yields:
```julia
julia> code_typed(LoopVectorization._avx_!, (argtypes_ssa[1], argtypes_ssa[2], argtypes_ssa[3], argtypes_ssa[4],argtypes_ssa[5],argtypes_ssa[6]))
Any[]

julia> code_typed(LoopVectorization._avx_!, (argtypes_ssa[1], argtypes_ssa[2], argtypes_ssa[3], argtypes_ssa[4],argtypes_ssa[5], acorrect))
1-element Vector{Any}:
```

timholy:
I don't really understand the rules for wrapping and unwrapping `Val`, either. @vchuravy, any insight?

vchuravy:
> This doesn't fixes the `QuoteNode`s of my motivating example, but not the `PartialStruct`, because in [unwrapconst](https://github.com/JuliaDebug/Cthulhu.jl/blob/028eb04e8b85e17164b5c0343686405331e75123/src/reflection.jl#L52):
> 
> ```julia
>         elseif isa(a, Core.Compiler.PartialStruct)
>             a = Core.Typeof(a.typ)
>         end
> ```
> 
> `a.typ` is the correct answer/makes it work, while `Core.Typeof(a.typ)` is incorrect for my example.
> 
> I don't know enough about the history of Julia internals nor their behavior in a particular snapshot to say what sort of special casing is needed here to get it to do `a.typ` in my case without breaking anything.

Yes that observation looks correct we should use `a.typ` directly. Can you add that to this PR?

vchuravy (approved):

chriselrod:
> This doesn't fixes the `QuoteNode`s of my motivating example, but not the `PartialStruct`

Oops, "doesn't" shouldn't have been there. Edited my comment.

> I don't really understand the rules for wrapping and unwrapping `Val`, either. @vchuravy, any insight?

It's surprising that making it a subtype changes the behavior:
```julia
julia> struct MyVal{N} end

julia> struct MyValInteger{N} <: Integer end

julia> @noinline function foo(::Union{MyVal{N},MyValInteger{N}}) where {N}
           i = 0 # code to make function + `@noinline` penalty big enough to not be inlined
           while i < N
               i += 1
           end
           i
       end

julia> add3_myval(x) = x + foo(MyVal{3}())
add3_myval (generic function with 1 method)

julia> add3_myvalinteger(x) = x + foo(MyValInteger{3}())
add3_myvalinteger (generic function with 1 method)

julia> @code_typed add3_myval(2)
CodeInfo(
1 ─ %1 = invoke Main.foo($(QuoteNode(MyVal{3}()))::MyVal{3})::Int64
│   %2 = Base.add_int(x, %1)::Int64
└──      return %2
) => Int64

julia> @code_typed add3_myvalinteger(2)
CodeInfo(
1 ─ %1 = invoke Main.foo(MyValInteger{3}()::MyValInteger{3})::Int64
│   %2 = Base.add_int(x, %1)::Int64
└──      return %2
) => Int64
```

> Yes that observation looks correct we should use `a.typ` directly. Can you add that to this PR?

Done.

timholy:
Thanks @chriselrod!

PR #126, opened 2021-02-16
Minimal switch to AbstractInterpreter

### LINKED ISSUE #124: Update to use AbstractInterpreter framework
Cthulhu should ideally be updated to use an `AbstractInterpreter` to do a full inference from the call site and then just be an interactive utility to explore what inference actually did. At the moment if inference does things like limiting call cycles, Cthulhu will re-do inference on descend and give a misleading view of the quality of type information. An additional benefit of using the AbstractInterpreter framework would be that Cthulhu could expose inference remarks. Perhaps @aviatesk would be interested in looking at this, since JET.jl is somewhat similar.

Keno:
This is the minimal work required to switch Cthulhu to AbstractInterpreter.
There are several things I would like to build on top of this:
    - Read callsites out of statement info rather than reconstructing it
    - Support for Opaque Closures
    - More consistent use of AbstractInterpreter throughout the API
    - Expose inference remarks

I would like to propose that we make a codebase break at Julia 1.6
for Cthulhu, moving the current master to a branch and requiring
Julia 1.7 on Cthulhu master. That gives us the rest of the 1.7 cycle
to update Cthulhu and figure out what the right interfaces between
Cthulhu and Base are.

Fixes #124

cc @maleadt

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,12 +13,6 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.0'
-          - '1.1'
-          - '1.2'
-          - '1.3'
-          - '1.4'
-          - '1.5'
           - 'nightly'
         os:
           - ubuntu-latest
--- a/Project.toml
+++ b/Project.toml
@@ -14,7 +14,7 @@ Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 [compat]
 CodeTracking = "0.5, 1"
 FoldingTrees = "1"
-julia = "1"
+julia = "1.7"
 
 [extras]
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -39,6 +39,7 @@ end
 """
 const CONFIG = CthulhuConfig()
 
+include("interpreter.jl")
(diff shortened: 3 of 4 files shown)


vchuravy:
> I would like to propose that we make a codebase break at Julia 1.6
> for Cthulhu, moving the current master to a branch and requiring
> Julia 1.7 on Cthulhu master. That gives us the rest of the 1.7 cycle
> to update Cthulhu and figure out what the right interfaces between
> Cthulhu and Base are.

Good by me.

aviatesk:
I'll take a look on this PR during this week.

Keno:
No worries, this PR is mostly the trial balloon to figure out what to do about version compats and such. Actually making use of AbstractInterpreter will take a couple more PRs - just wanted to keep you in the loop.

timholy:
Strongly supportive of this plan. As it stands, the inaccuracies in specialization heuristics alone would be worth the break.

Just a comment, our confidence in doing refactorings like this would be greatly enhanced by having a better test suite. Something for us all to aspire to. I've been trying to avoid fixing bugs or add functionality without accompanying tests, but we have a long ways to catch up.

Keno:
Alright, the plan here is:
- Branch a `16-compat` branch from current master
- Update this PR to update CI config to only run on master
- Merge this PR
- Start iterating from there by cleaning out old code/updating to use new interfaces, etc.

PR #128, opened 2021-02-17
WIP: Switch callsite detection to use stmt info

Keno:
Rips out all the heuristics that are duplicating the analysis
that inference already does to figure out what is getting called.
This should be much more precise and general. We may want to
revisit the whole CallInfo setup in the future to maybe use the
stmt info more directly, but this should be good for the moment.

Depends on https://github.com/JuliaLang/julia/pull/39716 (so WIP
until that is merged).

@vchuravy @timholy I'll need to do some more testing here, but you may want to run this through its paces to see if it catches all the use cases you have or if we need to make Base more verbose in some cases.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -150,6 +150,41 @@ const descend = descend_code_typed
 
 descend(ci::CthulhuInterpreter, mi::MethodInstance; kwargs...) = _descend(ci, mi; iswarn=false, interruptexc=false, kwargs...)
 
+function codeinst_rt(code::CodeInstance)
+    rettype = code.rettype
+    if isdefined(code, :rettype_const)
+        rettype_const = code.rettype_const
+        if isa(rettype_const, Vector{Any}) && !(Vector{Any} <: rettype)
+            return PartialStruct(rettype, rettype_const)
+        elseif rettype <: Core.OpaqueClosure && isa(rettype_const, PartialOpaque)
+            return rettype_const
+        elseif isa(rettype_const, InterConditional)
+            return rettype_const
+        else
+            return Const(rettype_const)
+        end
+    else
+        return rettype
+    end
+end
+
+function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool)
+    if !optimize
+        codeinf = copy(interp.unopt[mi].src)
+        rt = interp.unopt[mi].rt
+        infos = interp.unopt[mi].stmt_infos
+        slottypes = codeinf.slottypes
+    else
+        codeinst = interp.opt[mi]
+        ir = Core.Compiler.copy(codeinst.inferred.ir)
+        codeinf = ir
+        rt = codeinst_rt(codeinst)
+        infos = ir.stmts.info
+        slottypes = ir.argtypes
+    end
+    (codeinf, rt, infos, slottypes)
+end
+
 ##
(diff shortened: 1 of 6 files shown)


src/Cthulhu.jl
```diff
@@ -157,13 +166,28 @@ function _descend(interp::CthulhuInterpreter, mi::MethodInstance; iswarn::Bool,
...
 
-        display_CI && cthulu_typed(stdout, debuginfo_key, codeinf, rt, mi, iswarn, verbose)
-        display_CI = true
+        if optimize && interp.opt[mi].inferred === nothing
+            # This was inferred to a pure constant - we have no code to show,
+            # but make something up that looks plausible.
+            println(stdout)
+            println(stdout, "│ ─ $(string(Callsite(-1, MICallInfo(mi, interp.opt[mi].rettype), :invoke)))")
+            println(stdout, "│  return ", Const(interp.opt[mi].rettype_const))
+            println(stdout)
```
timholy, line 176:
```suggestion
            if display_CI
                println(stdout)
                println(stdout, "│ ─ $(string(Callsite(-1, MICallInfo(mi, interp.opt[mi].rettype), :invoke)))")
                println(stdout, "│  return ", Const(interp.opt[mi].rettype_const))
                println(stdout)
            end
```

src/reflection.jl
```diff
@@ -35,53 +36,63 @@ function transform(::Val{:CuFunction}, callsite, callexpr, CI, mi, slottypes; pa
...
+            if info !== nothing
+                rt = CI.ssavaluetypes[id]
+                was_return_type = false
+                if isa(info, Core.Compiler.ReturnTypeCallInfo)
+                    info = info.info
+                    was_return_type = true
+                end
+                if isa(info, MethodMatchInfo)
+                    if info.results === missing
+                        continue
```
timholy, line 56: In the early days should we `@warn`? We could suppress that before making a release, perhaps.

timholy (approved):
I can't really review the small CUDA elements, but overall this is very welcome!

timholy:
I haven't tested yet locally...I'm still running 1.6-rc1 to keep putting it through its paces in preparation for release. But perhaps it's time to transition to 1.7 and just download a binary for 1.6.

src/reflection.jl
Keno: This happens when there is too many applicable matches. That should also generate an entry in the inference msgs. Those aren't hooked up in the UI yet, but that's one of the next things I want to do.

vchuravy (approved):
The CUDA piece brings up the question of how to switch the underlying Abstract Interpreter...
I suspect that ideally we would switch and run Cthulhu over the AbstractInterpreter in GPUCompiler.jl

as side from that LGTM

Keno:
Now based on https://github.com/JuliaLang/julia/pull/39885.

Keno:
Alright, base should be done now. I'll wait for that to come through the nightlies and then plan to merge this.

timholy:
I'm super-excited about this, it's a much-needed strengthening of the foundations.

Keno:
Alright, this is merged :tada: . I'm sure various things are broken. I played with it on Diffractor and it worked fine (better than before because it now has precise tracking of things like constprop), but I don't use all parts of the Cthulhu functionality, so I'm sure things will need to be rebuilt.

PR #129, opened 2021-02-25
Fix a typo?

tkf:
This looks like a typo.


timholy:
Did you notice this just by reading or did it trigger an error? If so a test would be great if it's not too much to reproduce.

tkf:
Ah yes, good point. I should've noted that I just found it by reading it (when I was implementing #130). So, no, I don't know any case that triggers an error.

timholy:
OK, thanks again!

PR #137, opened 2021-03-12
Minimal fix for change in CostCallInfo

Keno:
This is just the general fix to try and get tests to pass again and Cthulhu somewhat working, I don't think it's optimal, but it's also not clear to me how to tell from the new design of ConstCallInfo, what to do if the result is `nothing`, i.e. what choices to present to the user there, so we may have to take another look at how this info is computed in Base. @aviatesk could you think about what needs to be done here?

### DIFF
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -76,7 +76,6 @@ function find_callsites(CI::Union{Core.CodeInfo, IRCode}, stmt_info::Union{Vecto
                     elseif isa(info, ConstCallInfo)
                         if length(info.results) == 1
                             result = info.results[1]
-                            @assert length(info.results) == 1
                             return [ConstPropCallInfo(MICallInfo(result.linfo, rt), result)]
                         else
                             vcat(process_info(info.call),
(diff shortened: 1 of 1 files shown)


src/reflection.jl
```diff
@@ -74,8 +74,15 @@ function find_callsites(CI::Union{Core.CodeInfo, IRCode}, stmt_info::Union{Vecto
                         # we ignore any implicit iterate calls.
                         return process_info(info.call)
                     elseif isa(info, ConstCallInfo)
-                        result = info.result
-                        return [ConstPropCallInfo(MICallInfo(result.linfo, rt), result)]
+                        if length(info.results) == 1
+                            result = info.results[1]
+                            @assert length(info.results) == 1
```
timholy, line 79:
```suggestion
```
The assert seems guaranteed to pass given the condition at the beginning of this block

Keno:
Merging this so people won't have a broken Cthulhu experience, but we should revisit this since I don't think it's quite correct.

aviatesk:
>  @aviatesk could you think about what needs to be done here?

Yeah I will look into this tmrw.

aviatesk:
#139

PR #140, opened 2021-03-14
minor code quality improvements

aviatesk:
Should not change any functionality, just improve type stabilities and removed one dead assignment.


aviatesk:
Let me merge this.

PR #144, opened 2021-03-20
fix #142

### LINKED ISSUE #142: doesn't seem to work on julia-master
```
ERROR: MethodError: no method matching argextype(::Core.SSAValue, ::Core.CodeInfo, ::Vector{Any}, ::Nothing)
Closest candidates are:
  argextype(::Any, ::Any, ::Vector{Any}) at compiler/utilities.jl:220
  argextype(::Any, ::Any, ::Vector{Any}, ::Vector{Any}) at compiler/utilities.jl:220
  argextype(::Any, ::Core.Compiler.IncrementalCompact, ::Vector{Any}) at compiler/ssair/queries.jl:93
  ...
Stacktrace:
 [1] find_callsites(CI::Core.CodeInfo, stmt_info::Vector{Any}, mi::Core.MethodInstance, slottypes::Nothing; params::Core.Compiler.NativeInterpreter, multichoose::Bool, kwargs::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:debuginfo,), Tuple{Symbol}}})
   @ Cthulhu ~/.julia/dev/Cthulhu/src/reflection.jl:50
 [2] _descend(interp::Cthulhu.CthulhuInterpreter, mi::Core.MethodInstance; override::Nothing, iswarn::Bool, params::Core.Compiler.NativeInterpreter, optimize::Bool, interruptexc::Bool, verbose::Bool, kwargs::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:debuginfo,), Tuple{Symbol}}})
   @ Cthulhu ~/.julia/dev/Cthulhu/src/Cthulhu.jl:237
 [3] _descend(interp::Cthulhu.CthulhuInterpreter, mi::Core.MethodInstance; override::Nothing, iswarn::Bool, params::Core.Compiler.NativeInterpreter, optimize::Bool, interruptexc::Bool, verbose::Bool, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Cthulhu ~/.julia/dev/Cthulhu/src/Cthulhu.jl:291
 [4] _descend(F::Any, TT::Any; params::Core.Compiler.NativeInterpreter, kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:iswarn,), Tuple{Bool}}})
   @ Cthulhu ~/.julia/dev/Cthulhu/src/Cthulhu.jl:369
 [5] _descend_with_error_handling(::Any, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, V, Tuple{Vararg{Symbol, N}}, NamedTuple{names, T}} where {V, N, names, T<:Tuple{Vararg{Any, N}}})
   @ Cthulhu ~/.julia/dev/Cthulhu/src/Cthulhu.jl:133
 [6] #descend_code_typed#46
   @ ~/.julia/dev/Cthulhu/src/Cthulhu.jl:105 [inlined]
 [7] descend_code_typed(f::Function, tt::Any)
   @ Cthulhu ~/.julia/dev/Cthulhu/src/Cthulhu.jl:105
 [8] top-level scope
   @ /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/InteractiveUtils/src/macros.jl:96
```

aviatesk:
EDIT: please see https://github.com/JuliaDebug/Cthulhu.jl/pull/144#issuecomment-812307339 for the actual changes this PR will make

By making sure `slottypes::Vector{Any}`

@vtjnash could you give me a MRE to reproduce #142 ? I'm still not sure why/how `CodeInfo.slottypes` can be `nothing`.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -6,6 +6,7 @@ using UUIDs
 
 using Core: MethodInstance
 const Compiler = Core.Compiler
+using Core.Compiler: MethodMatch, LimitedAccuracy, ignorelimited
 
 const mapany = Base.mapany
 
@@ -174,15 +175,18 @@ function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool)
         rt = interp.unopt[mi].rt
         infos = interp.unopt[mi].stmt_infos
         slottypes = codeinf.slottypes
+        if isnothing(slottypes)
+            slottypes = Any[ Any for i = 1:length(codeinf.slotflags) ]
+        end
     else
         codeinst = interp.opt[mi]
-        ir = Core.Compiler.copy(codeinst.inferred.ir)
+        ir = Core.Compiler.copy(codeinst.inferred.ir::IRCode)
         codeinf = ir
         rt = codeinst_rt(codeinst)
         infos = ir.stmts.info
         slottypes = ir.argtypes
     end
-    (codeinf, rt, infos, slottypes)
+    (codeinf, rt, infos, slottypes::Vector{Any})
 end
 
 ##
@@ -199,6 +203,7 @@ function _descend(interp::CthulhuInterpreter, mi::MethodInstance; override::Unio
     end
 
     display_CI = true
+    is_cached(key) = haskey(optimize ? interp.opt : interp.unopt, key)
     while true
         debuginfo_key = debuginfo ? :source : :none
 
@@ -214,25 +219,25 @@ function _descend(interp::CthulhuInterpreter, mi::MethodInstance; override::Unio
             callsites = Callsite[]
         else
             if override !== nothing
-                if !haskey(interp.unopt, override)
-                    Core.eval(Core.Main, :(the_issue = $override))
-                end
                 codeinf = optimize ? override.src : interp.unopt[override].src
                 rt = optimize ? override.result : interp.unopt[override].rt
                 if optimize
-                    codeinf = Core.Compiler.copy(codeinf.ir)
+                    codeinf = Core.Compiler.copy(codeinf.ir::IRCode)
                     infos = codeinf.stmts.info
                     slottypes = codeinf.argtypes
                 else
                     codeinf = copy(codeinf)
                     infos = interp.unopt[override].stmt_infos
                     slottypes = codeinf.slottypes
+                    if isnothing(slottypes)
+                        slottypes = Any[ Any for i = 1:length(codeinf.slotflags) ]
+                    end
                 end
             else
                 (codeinf, rt, infos, slottypes) = lookup(interp, mi, optimize)
             end
             preprocess_ci!(codeinf, mi, optimize, CONFIG)
-            callsites = find_callsites(codeinf, infos, mi, slottypes; params=params, kwargs...)
+            callsites = find_callsites(interp, codeinf, infos, mi, slottypes, optimize; params, kwargs...)
 
             display_CI && cthulu_typed(stdout, debuginfo_key, codeinf, rt, mi, iswarn, verbose)
             display_CI = true
@@ -277,7 +282,18 @@ function _descend(interp::CthulhuInterpreter, mi::MethodInstance; override::Unio
                 callsite = sub_callsites[cid]
             end
 
-            if info isa GeneratedCallInfo || info isa FailedCallInfo
+            if info isa UncachedCallInfo
+                @warn "Cthulhu can't recurse into the uncached callsite"
+                continue
+                # # or, we can forcibly enter into the frame, which the native interpreter threw away
+                # next_interp = CthulhuInterpreter()
+                # next_mi = get_mi(callsite.info)
+                # do_typeinf!(next_interp, next_mi)
+                # _descend(next_interp, next_mi;
+                #          params, optimize, iswarn, debuginfo=debuginfo_key, interruptexc, verbose, kwargs...)
+            end
+
+            if info isa GeneratedCallInfo || callsite.info isa FailedCallInfo
                 @error "Callsite %$(callsite.id) failed to be extracted" callsite
             end
 
@@ -298,6 +314,11 @@ function _descend(interp::CthulhuInterpreter, mi::MethodInstance; override::Unio
             verbose ⊻= true
         elseif toggle === :optimize
             optimize ⊻= true
+            if !is_cached(mi)
+                @warn "can't switch to post-optimization state, since this inference frame isn't cached"
+                optimize ⊻= true
+            end
+            continue
         elseif toggle === :debuginfo
             debuginfo ⊻= true
         elseif toggle === :highlighter
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -279,18 +279,19 @@ function _descend(interp::CthulhuInterpreter, mi::MethodInstance; override::Unio
                 if cid == -1
                     interruptexc ? throw(InterruptException()) : break
                 end
+
                 callsite = sub_callsites[cid]
+                info = callsite.info
             end
 
             if info isa UncachedCallInfo
-                @warn "Cthulhu can't recurse into the uncached callsite"
+                # forcibly enter and inspect the frame, which the native interpreter threw away
+                next_interp = CthulhuInterpreter()
+                next_mi = get_mi(info)
+                do_typeinf!(next_interp, next_mi)
+                _descend(next_interp, next_mi;
+                         params, optimize, iswarn, debuginfo=debuginfo_key, interruptexc, verbose, kwargs...)
                 continue
-                # # or, we can forcibly enter into the frame, which the native interpreter threw away
-                # next_interp = CthulhuInterpreter()
-                # next_mi = get_mi(callsite.info)
-                # do_typeinf!(next_interp, next_mi)
-                # _descend(next_interp, next_mi;
-                #          params, optimize, iswarn, debuginfo=debuginfo_key, interruptexc, verbose, kwargs...)
             end
 
             if info isa GeneratedCallInfo || callsite.info isa FailedCallInfo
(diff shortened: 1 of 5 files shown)


vtjnash:
sorry, this isn't a great example, since it is quite slow (and seems like Cthulhu no longer has the ability to cache between calls), but this is what I was trying to debug:
```
julia> using MixedModels: dataset                                                                
                                                                                                                                                                                                   
julia> using MixedModels                                                                                                                                                                           
                                                                                                 
julia> form = @formula((incid/hsz) ~ 1 + period + (1|herd))                                      
FormulaTerm                                                                                                                                                                                        
Response:                                                                                        
  (incid,hsz)->incid / hsz                                                                                                                                                                         Predictors:                                                                                                                                                                                          1                                   
  period(unknown)                     
  (herd)->1 | herd                    
                                                
julia> @descend GeneralizedLinearMixedModel(form, cbpp,                                          
                                                  Binomial();                                    
                                                  wts=float(cbpp.hsz))
> select 1
ERROR: KeyError: key MethodInstance for (::Core.var"#Type##kw")(::NamedTuple{(:wts,), Tuple{Vector{Float64}}}, ::Type{GeneralizedLinearMixedModel}, ::StatsModels.FormulaTerm{StatsModels.FunctionTerm{typeof(/), var"#3#5", (:incid, :hsz)}, Tuple{StatsModels.ConstantTerm{Int64}, StatsModels.Term, StatsModels.FunctionTerm{typeof(|), var"#4#6", (:herd,)}}}, ::Arrow.Table, ::Binomial{Float64}, ::LogitLink) not found

> select o
> select 2
ERROR: MethodError: no method matching argextype(::Core.SSAValue, ::Core.CodeInfo, ::Vector{Any}, ::Nothing)
```

aviatesk:
Thanks, and now the example will produce the same error as reported in https://github.com/JuliaLang/julia/issues/40048 ...
```julia
julia> using MixedModels: dataset; using MixedModels; form = @formula((incid/hsz) ~ 1 + period + (1|herd)); cbpp = dataset(:cbpp);

julia> @descend GeneralizedLinearMixedModel(form, cbpp,                                          
                                                  Binomial();                                    
                                                  wts=float(cbpp.hsz))
ERROR: StackOverflowError:
...
```

simeonschaub:
A more self-contained example:
```julia
julia> using Cthulhu

julia> @descend mapreduce(identity, vcat, ([5], [1.]))

│ ─ %-1  = invoke mapreduce(::typeof(identity),::typeof(vcat),::Tuple{Vector{Int64}, Vector{Float64}})::AbstractVector{T} where T
289 1 ─ %1 = invoke Base.:(var"#mapreduce#237")($(QuoteNode(Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}()))::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, _1::typeof(mapreduce), _2::Function, _3::Function, _4::Tuple{Vector{Int64}, Vector{Float64}})::AbstractVector{T} where T
    └──      return %1                                                                                          │

 • %1  = invoke #mapreduce#237(…,…,…,…,…)::AbstractVector{T} where T
   ↩
ERROR: KeyError: key MethodInstance for Base.var"#mapreduce#237"(::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, ::typeof(mapreduce), ::typeof(identity), ::typeof(vcat), ::Tuple{Vector{Int64}, Vector{Float64}}) not found
Stacktrace:
 [1] getindex
   @ ./dict.jl:482 [inlined]
 [2] _descend(interp::Cthulhu.CthulhuInterpreter, mi::Core.MethodInstance; override::Nothing, iswarn::Bool, params::Core.Compiler.NativeInterpreter, optimize::Bool, interruptexc::Bool, verbose::Bool, kwargs::Base.Pairs{Symbol, Symbol, Tuple{Symbol}, NamedTuple{(:debuginfo,), Tuple{Symbol}}})
   @ Cthulhu ~/.julia/packages/Cthulhu/7xpUc/src/Cthulhu.jl:208
 [3] _descend(interp::Cthulhu.CthulhuInterpreter, mi::Core.MethodInstance; override::Nothing, iswarn::Bool, params::Core.Compiler.NativeInterpreter, optimize::Bool, interruptexc::Bool, verbose::Bool, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Cthulhu ~/.julia/packages/Cthulhu/7xpUc/src/Cthulhu.jl:295
 [4] _descend(F::Any, TT::Any; params::Core.Compiler.NativeInterpreter, kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:iswarn,), Tuple{Bool}}})
   @ Cthulhu ~/.julia/packages/Cthulhu/7xpUc/src/Cthulhu.jl:373
 [5] _descend_with_error_handling(::Any, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, V, Tuple{Vararg{Symbol, N}}, NamedTuple{names, T}} where {V, N, names, T<:Tuple{Vararg{Any, N}}})
   @ Cthulhu ~/.julia/packages/Cthulhu/7xpUc/src/Cthulhu.jl:133
 [6] #descend_code_typed#46
   @ ~/.julia/packages/Cthulhu/7xpUc/src/Cthulhu.jl:105 [inlined]
 [7] descend_code_typed(f::Function, tt::Any)
   @ Cthulhu ~/.julia/packages/Cthulhu/7xpUc/src/Cthulhu.jl:105
 [8] top-level scope
   @ REPL[2]:1
```
Doesn't seem like this PR fixes this particular example though.

src/Cthulhu.jl
```diff
@@ -174,15 +174,18 @@ function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool)
         rt = interp.unopt[mi].rt
         infos = interp.unopt[mi].stmt_infos
         slottypes = codeinf.slottypes
+        if isnothing(slottypes)
+            slottypes = Any[ Any for i = 1:nslots ]
```
simeonschaub, line 178: `nslots` does not seem to be defined anywhere

aviatesk: ops I just forgot to push the local fix. Thanks.

vchuravy:
Great work! Is it possible to write a test for this!

aviatesk:
Okay, so the problem was kinda deeply nested than I expected.

The first thing to notice is that there are cases when we can't find a
cache in `CthulhuInterpreter.opt`.
AFAIU this is mainly because of limited frames; the native interpreter
doesn't cache them but there are still statement infos about them and
so the problem happens if a Cthulhu user tries to inspect it.

When trying to inspect uncached frames, the next problem is how to
handle them at pre-optimization phase. The `slottypes` of them would be
`nothing` at that point, and so we need to do something like
[what the external `OptimizationState` constructor does](https://github.com/JuliaLang/julia/blob/878e1cd52ebf14e8666d8d54a54a95ccee4405d3/base/compiler/optimize.jl#L69-L71).
The another problem is the return types of those frames can be annotated
as `LimitedAccuracy` at that point, and it contains recursive frame chain
and thus the printer can fail into infinite loop. The same issue happens
when Cthulhu tries to show `CodeInfo` object, which may also contain
`LimitedAccuracy`.

Now this commit addresses the issues outlined above:
- check if a callsite is cached or not, and wrap it into `UncachedCallInfo` if not,
  and Cthulhu doesn't recurse into the `UncachedCallInfo` (or we can even
  forcibly recurse into, but I decided not to do that, since it's really
  different behaviour from what the native interpreter did).
  Note that `UncachedCallInfo` won't happen when working on a pre-optimization state,
  since Cthulhu will cache unoptimized frames by hooking into
  `finish(::InferenceState, ::CthulhuInterpreter)` in that case,
  which caches a frame no matter whether it is limited or not.
- when working on a pre-optimization state, if the return type of a callsite
  is annotated as `LimitedAccuracy`, we widen it and wrap it into
`LimitedCallInfo`.
  When printing `LimitedCallInfo`, Cthulhu only indicates the call is limited,
  but won't try to print the whole `LimitedAccuracy` object.
- when working on pre-optimization state, we make sure we don't print `LimitedAccuracy`
  included in `CodeInfo`. Currently I wrote an naive hack to just copy
  the original `CodeInfo` and widen `LimitedAccuracy`s in `ssavaluetypes` and `rettype`.

The changes above will fix #142 as well as [the case @simeonschaub reporeted here](https://github.com/JuliaDebug/Cthulhu.jl/pull/144#issuecomment-810959321).

---

@Keno (@vtjnash) do these changes sound fine ?

vchuravy:
Thanks @aviatesk! Can you rebase?

aviatesk:
Done :)

test/runtests.jl
```diff
@@ -118,9 +123,30 @@ let callsites = find_callsites_by_ftt(f_matches, Tuple{Any, Any}; optimize=false
     @test callinfo isa Cthulhu.MultiCallInfo
 end
 
-@testset "union-split constant-prop'ed callsites" begin
-    anonymous_module() = Core.eval(@__MODULE__, :(module $(gensym()) end))::Module
+# NOTE some inference won't work with inner functions
+anonymous_module() = Core.eval(@__MODULE__, :(module $(gensym()) end))::Module
```
simeonschaub, line 127: Just FYI, I think you can just use `Module()` for this, but doesn't really matter here.

aviatesk: That's neat. Thanks.

src/Cthulhu.jl
```diff
@@ -277,7 +282,18 @@ function _descend(interp::CthulhuInterpreter, mi::MethodInstance; override::Unio
                 callsite = sub_callsites[cid]
             end
 
-            if info isa GeneratedCallInfo || info isa FailedCallInfo
+            if info isa UncachedCallInfo
+                @warn "Cthulhu can't recurse into the uncached callsite"
+                continue
+                # # or, we can forcibly enter into the frame, which the native interpreter threw away
```
vchuravy, line 288: I suspect we would want to do that. Otherwise it will be hard for the user to investigate further from this point.

aviatesk: Yeah, that's also a valid argument. Applied with some other minor fixes: 9624b3b

aviatesk:
Let me merge this within a day.

aviatesk:
Going to merge.

PR #146, opened 2021-03-30
make IRCode printing respect debuginfo

### LINKED ISSUE #143: debuginfo info option broken/inconsistent
We seem to have lost debuginfo formatting on master (still works for some options, but optimized IR is broken)

simeonschaub:
The function names on the right are still shown, even if
`debuginfo=:none`, but now `verbose_linetable` is used if
`debuginfo=:source`.

Closes #143.


simeonschaub:
OK if I merge this?

PR #160, opened 2021-06-23
add option to print inlining costs

simeonschaub:
This allows printing inlining costs alongside optimized IR similar to
`Base.print_statement_costs`, which I found useful for debugging why
some functions are not inlined. They are highlighted green, because I
found them hard to distinguish otherwise, but I am open to suggestions
regarding formatting. This relies on JuliaLang/julia#41257, which I hope
can be backported to 1.7, so we don't have to worry about compatibility.

### DIFF
--- a/src/codeview.jl
+++ b/src/codeview.jl
@@ -72,53 +72,65 @@ function is_type_unstable(code::Union{IRCode, CodeInfo}, idx::Int, used::BitSet)
 end
 
 cthulhu_warntype(args...) = cthulhu_warntype(stdout, args...)
-function cthulhu_warntype(io::IO, src, rettype, debuginfo, stable_code)
-    debuginfo = IRShow.debuginfo(debuginfo)
-    lineprinter = __debuginfo[debuginfo]
-    lambda_io::IOContext = io
-    if hasfield(typeof(src), :slotnames) && src.slotnames !== nothing
-        slotnames = Base.sourceinfo_slotnames(src)
-        lambda_io = IOContext(lambda_io, :SOURCE_SLOTNAMES => slotnames)
-        show_variables(io, src, slotnames)
+function cthulhu_warntype(io::IO, src, rettype, debuginfo, stable_code, inline_cost=false)
+    if inline_cost
+        error("Need a MethodInstance to show inlining costs. Call `cthulhu_typed` directly instead.")
     end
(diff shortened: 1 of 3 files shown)


simeonschaub:
An example of what this looks like:
Screenshot_2021-06-23_14-38-17

aviatesk (approved):
Nice feature !

src/codeview.jl
```diff
@@ -72,53 +72,65 @@ function is_type_unstable(code::Union{IRCode, CodeInfo}, idx::Int, used::BitSet)
...
+        println(io)
     else
-        bb_color = (src isa IRCode && debuginfo === :compact) ? :normal : :light_black
-        irshow_config = IRShowConfig(lineprinter(src); bb_color)
-        show_ir(io, src, irshow_config)
+        println(io, "│ ─ $(string(Callsite(-1, MICallInfo(mi, rettype), :invoke)))")
     end
-    println(io)
+
+    if src isa IRCode && inline_cost
```
aviatesk, line 110: We can just allow the inlining cost printing for unoptimized source as well ?

```suggestion
    if inline_cost
```

simeonschaub: I did actually experiment with that, but the issue is that any call that isn't an intrinsic is assumed to be a dynamic call, so it's probably more misleading than useful. I wonder whether we can get the recursive total costs after inlining for each call from the AbstractInterpreter, but that is probably a separate PR.

aviatesk: Ah, I see.
For now maybe it'd be more kind if we print some warning message like:
```juila
        elseif toggle === :optimize
            optimize ⊻= true
            if !is_cached(mi)
                @warn "can't switch to post-optimization state, since this inference frame isn't cached"
                optimize ⊻= true
            end
            continue
        elseif toggle === :debuginfo
            debuginfo = DebugInfo((Int(debuginfo) + 1) % 3)
        elseif toggle === :inline_cost
            if !optimize
                @warn "enable optimization to see the inlining costs"
            end
            inline_cost ⊻= true
```
?

simeonschaub: Yeah, that sounds reasonable. Do you just want to push that to this branch?

aviatesk: Okay, I will push the change and merge this.

PR #161, opened 2021-06-27
Allow _descend on signature tuple type

Keno:
Makes it easier to decend if you can't (or its hard to) construct
an object of the function type, but you do have the signature.

Also, while I'm at it, fix a 1.7 compatibility issue.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -376,22 +376,30 @@ function do_typeinf!(interp, mi)
     return nothing
 end
 
-function get_specialization(@nospecialize(F), @nospecialize(TT))
-    sigt = Base.signature_type(F, TT)
-    match = Base._which(sigt)
+function get_specialization(@nospecialize(TT))
+    match = Base._which(TT)
     mi = Core.Compiler.specialize_method(match)
     return mi
 end
 
-function mkinterp(@nospecialize(F), @nospecialize(TT))
+function get_specialization(@nospecialize(F), @nospecialize(TT))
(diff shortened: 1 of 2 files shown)


src/reflection.jl
```diff
@@ -179,7 +179,11 @@ function find_callsites(interp::CthulhuInterpreter, CI::Union{Core.CodeInfo, IRC
 end
 
 function dce!(ci, mi)
-    argtypes = Core.Compiler.matching_cache_argtypes(mi, nothing)[1]
+    if VERSION >= v"1.7.0-DEV.705"
```
simeonschaub, line 182: Do we even need the check here? I thought the goal was to just require Julia 1.7.0 for the next Cthulhu version.

Keno: Probably. This helps for now though. We can remove it once we bump the dependency.

src/Cthulhu.jl
```diff
@@ -395,6 +398,11 @@ function _descend(@nospecialize(F), @nospecialize(TT); params=current_params(),
     _descend(interp, mi; params=params, kwargs...)
 end
 
+function _descend(@nospecialize(TT::Type{<:Tuple}); params=current_params(), kwargs...)
```
aviatesk, line 401: We can unify these two definitions of `_descend` by:
```julia
function _descend(@nospecialize(args...); params=current_params(), kwargs...)
    (interp, mi) = mkinterp(args...)
    _descend(interp, mi; params=params, kwargs...)
end
```

aviatesk (approved):

PR #172, opened 2021-07-23
enable descend into `Task`

aviatesk:
Just found we can't recurse into `Task` closure v2..


aviatesk:
@vchuravy going to merge, let me know if I'm doing wrong.

PR #176, opened 2021-07-29
Fix is_callsite for WrappedCallInfos

timholy:
This also:
- ensures that the DCE test runs on CI
- marks the DCE tests as broken (is this a Julia bug?)
- adds a `.gitignore`

Here's the DCE result:
```julia
julia> Base.@propagate_inbounds function f(x)
           @boundscheck error()
       end

julia> g(x) = @inbounds f(x)
g (generic function with 1 method)

julia> h(x) = f(x)
h (generic function with 1 method)

julia> function process(@nospecialize(f), @nospecialize(TT); optimize=true)
           (interp, mi) = Cthulhu.mkinterp(f, TT)
           (ci, rt, infos, slottypes) = Cthulhu.lookup(interp, mi, optimize)
           Cthulhu.preprocess_ci!(ci, mi, optimize, Cthulhu.CthulhuConfig(dead_code_elimination=true))
           interp, ci, infos, mi, rt, slottypes
       end
process (generic function with 1 method)

julia> (_,CI, _, _, _, _) = process(g, Tuple{Vector{Float64}});

julia> CI
1 1 ─     goto #3 if not false                                                                                                                                                                      │╻ f
  2 ─     invoke Main.error()::Union{}                                                                                                                                                              ││
  └──     unreachable                                                                                                                                                                               ││
  3 ─     goto #4                                                                                                                                                                                   ││
  4 ─     return nothing                                                                                                                                                                            │ 
```
I would have thought that more of the `goto`s would have been resolved.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,6 +38,9 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
+      - name: Bounds check
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: julia --project -e 'using Pkg; Pkg.test("Cthulhu"; coverage=true, julia_args=["--check-bounds=no"])'
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v1
         with:
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+Manifest.toml
+*.jl.cov
+*.jl.*.cov
+*.jl.mem
--- a/src/callsite.jl
+++ b/src/callsite.jl
@@ -270,8 +270,7 @@ _wrapped_callinfo(limiter, ::UncachedCallInfo) = print(limiter, "uncached")
 
 is_callsite(cs::Callsite, mi::MethodInstance) = is_callsite(cs.info, mi)
 is_callsite(info::MICallInfo, mi::MethodInstance) = get_mi(info) === mi
-is_callsite(info::LimitedCallInfo, mi::MethodInstance) = is_callsite(info.ci, mi)
-is_callsite(info::UncachedCallInfo, mi::MethodInstance) = is_callsite(info.ci, mi)
+is_callsite(info::WrappedCallInfo, mi::MethodInstance) = is_callsite(get_wrapped(info), mi)
 is_callsite(info::ConstPropCallInfo, mi::MethodInstance) = is_callsite(info.mi, mi)
 is_callsite(info::TaskCallInfo, mi::MethodInstance) = is_callsite(info.ci, mi)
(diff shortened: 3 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -38,6 +38,8 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
+      - name: Bounds check
+        run: julia -e 'using Pkg; Pkg.test("Cthulhu"; julia_args=["--check-bounds=no"])
```
aviatesk, line 42:
```suggestion
        run: julia -e 'using Pkg; Pkg.test("Cthulhu"; julia_args=["--check-bounds=no"])'
```

vchuravy:
> I would have thought that more of the gotos would have been resolved.

What was the code before process? We used to run DCE twice to cleanup more code.

timholy:
```julia
julia> code_typed(g, Tuple{Vector{Float64}})
1-element Vector{Any}:
 CodeInfo(
1 ─     goto #3 if not false
2 ─     invoke Main.error()::Union{}
└──     unreachable
3 ─     goto #4
4 ─     return nothing
) => Nothing

julia> code_typed(g, Tuple{Vector{Float64}}; optimize=false)
1-element Vector{Any}:
 CodeInfo(
1 ─     $(Expr(:inbounds, true))::Any
│       (val = Main.f(x))::Core.Const(nothing)
│       $(Expr(:inbounds, :pop))::Any
└──     return val
) => Nothing
```
It's calling `dce!` twice from here: https://github.com/JuliaDebug/Cthulhu.jl/blob/ed2d80a142a8c2d3a5dbdba444111e753ee2f583/src/reflection.jl#L210-L218

aviatesk:
This diff will "fix" the broken tests:
```diff
diff --git a/test/runtests.jl b/test/runtests.jl
index a8114fd..4aeef98 100644
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -25,7 +25,7 @@ end
 function process(@nospecialize(f), @nospecialize(TT); optimize=true)
     (interp, mi) = Cthulhu.mkinterp(f, TT)
     (ci, rt, infos, slottypes) = Cthulhu.lookup(interp, mi, optimize)
-    Cthulhu.preprocess_ci!(ci, mi, optimize, Cthulhu.CthulhuConfig(dead_code_elimination=true))
+    ci = Cthulhu.preprocess_ci!(ci, mi, optimize, Cthulhu.CthulhuConfig(dead_code_elimination=true))
     interp, ci, infos, mi, rt, slottypes
 end
```

And this suggests, `preprocess!` doesn't change the input argument.

aviatesk:
https://github.com/JuliaDebug/Cthulhu.jl/blob/cffba70692bcd065165c40af00c7e53427fa529d/src/reflection.jl#L197 
will create new `IRCode`.

So I guess we need to `x = preprocess_ci!` everywhere actually ?

timholy:
Thanks @aviatesk, see #177

PR #177, opened 2021-07-29
Fix DCE

### LINKED ISSUE #153: code printing contains numerous mistakes in numbering?
On https://github.com/vtjnash/julia/pull/new/jn/multireplacer commit a5b3b28381246ce38942a84781ec5776db219993, I noticed when testing the new code with `@descend optimize=true replace("s", 'c'=>"BOOM");` that descending into this function shows numerous incorrect SSAValues of the form:
```
85 ── %171 = goto #101
```

timholy:
Fixes #153 (probably, CC @vtjnash)

Co-authored-by: Shuhei Kadowaki

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,7 +40,7 @@ jobs:
       - uses: julia-actions/julia-runtest@v1
       - name: Bounds check
         if: ${{ matrix.os == 'ubuntu-latest' }}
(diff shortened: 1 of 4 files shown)


aviatesk (approved):
Ahh, `:code_coverage_effect` stuff is really tricky.

timholy:
It's also a pity we don't have `Pkg.test("Cthulhu"; checkbounds=nothing)` to prevent `Pkg.test` from setting `--check-bounds`. The default gives a value of 0, `--check-bounds=yes` gives 1, and `--check-bounds=no` gives 2, so AFAICT using `Pkg.test` there's no way to get 0.

PR #179, opened 2021-07-29
Handle a few corner cases

timholy:


timholy:
Hah. The lines I added show up as untested, because they were designed to handle an empty function body. But when run under `--code-coverage`, no function has an empty body. I just added a third run of the test suite with coverage off. Seem OK, or are people worried about growing CI times?

vchuravy:
Thanks Tim!

PR #180, opened 2021-07-30
Fix handling of MethodResultPure

timholy:
Formerly Cthulhu tried to ignore statements whose `info` was
 `MethodResultPure` . This was marked as an annotation TODO
 but also didn't quite work:

```julia
julia> using Cthulhu, StaticArrays

# Using `find_callsites_by_ftt from the tests
julia> callsites = find_callsites_by_ftt(iterate, Tuple{SVector{3,Int}, Tuple{SOneTo{3}}}; optimize=false)
CI = CodeInfo(
    @ abstractarray.jl:1142 within `iterate`
1 ─       (y = Core._apply_iterate(Base.iterate, Base.iterate, state))::Core.Const((1, 1))
│   @ abstractarray.jl:1143 within `iterate`
│   %2  = (y::Core.Const((1, 1)) === Base.nothing)::Core.Const(false)
└──       goto #2 if not %2
    @ abstractarray.jl:1144 within `iterate`
2 ─ %4  = Base.getindex(y::Core.Const((1, 1)), 1)::Core.Const(1)
│   %5  = Base.getindex(A, %4)::Int64
│   %6  = Base.getindex(state, 1)::Core.Const(SOneTo(3))
│   %7  = Core.tuple(%6)::Core.Const((SOneTo(3),))
│   %8  = Base.tail(y::Core.Const((1, 1)))::Core.Const((1,))
│   %9  = Core._apply_iterate(Base.iterate, Core.tuple, %7, %8)::Core.Const((SOneTo(3), 1))
│   %10 = Core.tuple(%5, %9)::Core.PartialStruct(Tuple{Int64, Tuple{SOneTo{3}, Int64}}, Any[Int64, Core.Const((SOneTo(3), 1))])
└──       return %10
)
c = :(_4 = Core._apply_iterate(Base.iterate, Base.iterate, _3))
info = Core.Compiler.MethodResultPure(Core.Compiler.MethodMatchInfo(Core.Compiler.MethodLookupResult(Any[Core.MethodMatch(Tuple{typeof(iterate), SOneTo{3}}, svec(3), iterate(::SOneTo{n}) where n in StaticArrays at /home/tim/.julia/packages/StaticArrays/gBeIF/src/SOneTo.jl:41, true)], Core.Compiler.WorldRange(0x00000000000079d9, 0xffffffffffffffff), false)))
```
There error here comes from https://github.com/JuliaDebug/Cthulhu.jl/blob/349fb73e0b701d2bfb014224c19227c59d6b745b/src/reflection.jl#L106-L109

It turns out this happens because the compiler introduced a `MethodResultPure`
wrapped in a `UnionSplitApplyCallInfo`, and our logic for bailing out from pure
happened outside of the `mapreduce` call used in handling `UnionSplitApplyCallInfo`.

This fixes it by defining a new `WrappedCallInfo` subtype for marking these `info`s.
On this branch, here's what you now get:
```
julia> callsites = find_callsites_by_ftt(iterate, Tuple{SVector{3,Int}, Tuple{SOneTo{3}}}; optimize=false)
5-element Vector{Cthulhu.Callsite}:
 %1  = < pure, uncached > iterate(::SOneTo{3})::Core.Const((1, 1))
...
```

### DIFF
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -51,11 +51,16 @@ function find_callsites(interp::CthulhuInterpreter, CI::Union{Core.CodeInfo, IRC
             info = stmt_info[id]
             if info !== nothing
                 rt = argextype(SSAValue(id), CI, sptypes, slottypes)
-                was_return_type = false
-                if isa(info, MethodResultPure)
-                    # TODO: We could annotate this in the UI
-                    continue
+                # in unoptimized IR, there may be `slot = rhs` expressions, which `argextype` doesn't handle
+                # so extract rhs for such an case
+                args = c.args
+                if !optimize
+                    if isexpr(c, :(=))
+                        args = c.args[2].args
+                    end
                 end
+                types = mapany(arg -> widenconst(argextype(arg, CI, sptypes, slottypes)), args)
+                was_return_type = false
                 if isa(info, Core.Compiler.ReturnTypeCallInfo)
                     info = info.info
                     was_return_type = true
@@ -73,6 +78,8 @@ function find_callsites(interp::CthulhuInterpreter, CI::Union{Core.CodeInfo, IRC
                             mici = MICallInfo(mi, rt)
                             return is_cached(mi) ? mici : UncachedCallInfo(mici)
                         end
+                    elseif isa(info, MethodResultPure)
+                        return Any[PureCallInfo(types, rt)]
                     elseif isa(info, UnionSplitInfo)
                         return mapreduce(process_info, vcat, info.matches)
                     elseif isa(info, UnionSplitApplyCallInfo)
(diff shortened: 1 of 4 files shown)


src/reflection.jl
```diff
@@ -73,6 +69,8 @@ function find_callsites(interp::CthulhuInterpreter, CI::Union{Core.CodeInfo, IRC
                             mici = MICallInfo(mi, rt)
                             return is_cached(mi) ? mici : UncachedCallInfo(mici)
                         end
+                    elseif isa(info, MethodResultPure)
+                        return Any[PureCallInfo(only(process_info(info.info)))]
```
aviatesk, line 73: I agree with moving the handling of `MethodResultPure` into `process_info`, but I don't think `PureCallInfo` should be subtype of `WrappedCallInfo`.
`MethodResultPure` often doesn't wrap other call infos (e.g. see [here](https://github.com/JuliaLang/julia/blob/795935fd3a2d97b2f948cfb82a18da48743b622d/base/compiler/abstractinterpretation.jl#L1291)), and passing such cases can result in error:
```julia
julia> @eval descend(; optimize=false) do
           length($(Core.svec(0,1,2)))
       end
ERROR: ArgumentError: Collection is empty, must contain exactly 1 element
```
, and also it might be misleading to print "uncached" (i.e. when an user annotated a method as "pure" https://github.com/JuliaLang/julia/blob/795935fd3a2d97b2f948cfb82a18da48743b622d/base/compiler/abstractinterpretation.jl#L60) because `MethodResultPure` means super aggressive constant propagation and it's really opposite to "deoptimization", which "uncached" callsites usually mean.

timholy:
Updated with the excellent additions from @aviatesk merged into this branch in #182.

aviatesk:
Thanks, Tim.

PR #181, opened 2021-07-30
Fix printing bugs

timholy:
There was some nice functionality built into `TextWidthLimiter`,
but there was a lot that didn't quite work as expected due to a
lack of tests. This adds new tests, fixes bugs, and introduces
two changes to printing:
- print `::Vararg{Char}` as `::Char...` (with a more widely spaced
  three-dots than used for the "omission" marker
- use "::…" as the return-type annotation when there isn't space
  for the full thing but for which there is room for 3 characters.

If folks don't like these changes, I can revert them.

### DIFF
--- a/src/callsite.jl
+++ b/src/callsite.jl
@@ -115,6 +115,7 @@ end
 TextWidthLimiter(io::IO, limit) = TextWidthLimiter(io, 0, limit)
 has_space(limiter::TextWidthLimiter, width::Int) = limiter.width + width < limiter.limit - 1
 has_space(limiter::TextWidthLimiter, s) = has_space(limiter, textwidth(string(s)))
+has_space(::IO, s) = true
 function Base.print(io::TextWidthLimiter, s::String)
     io.width == io.limit && return 0
     width = textwidth(s::String)
@@ -126,15 +127,30 @@ function Base.print(io::TextWidthLimiter, s::String)
         for c in graphemes(s)
             cwidth = textwidth(c)
             if has_space(io, cwidth)
-                print(io, c)
+                print(io.io, c)
                 io.width += cwidth
             else
                 break
             end
         end
-        print(io, '…')
+        print(io.io, '…')
         io.width += 1
     end
+end
+
+function Base.print(io::TextWidthLimiter, c::Char)
+    tw = textwidth(c)
+    if has_space(io, tw)
+        print(io.io, c)
+        io.width += tw
+        return
+    end
+    return
+end
+
+function Base.take!(io::TextWidthLimiter)
+    io.width = 0
+    return take!(io.io)
 end
 
 function headstring(@nospecialize(T))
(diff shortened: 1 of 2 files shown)


src/callsite.jl
```diff
@@ -126,36 +127,53 @@ function Base.print(io::TextWidthLimiter, s::String)
...
         return headstring(Base.unwrap_unionall(T))
     else
-        return string(T.name)
+        return string(T.name.name)
     end
 end
 
 
 function __show_limited(limiter, name, tt, rt)
+    vastring(@nospecialize(T)) = Base.isvarargtype(T) ? headstring(T)*"..." : string(T)
```
simeonschaub, line 169: I think we will probably need to handle the case where `N` is also specified separately, i.e. something like `::Vararg{T, 3}`.

timholy: I think(?) those get expanded?

```julia
julia> getidx(A::Array{T}, i::Vararg{Integer,N}) where {T,N} = sin(A[i...])
getidx (generic function with 1 method)
julia> callsites = find_callsites_by_ftt(getidx, Tuple{Matrix{Bool}, Vararg{T,2} where T<:Integer}, optimize=false)
2-element Vector{Cthulhu.Callsite}:
 %2  = call #_apply_iterate(::#iterate{…},::#getindex{…},::Tuple{…},::Tuple{…})
 %3  = call sin(::Bool)::Float64

julia> cs = callsites[1]
%2  = call #_apply_iterate(::typeof(iterate),::typeof(getindex),::Tuple{Matrix{Bool}},::Tuple{Integer, Integer})::Bool
```

src/callsite.jl
```diff
@@ -115,6 +115,7 @@ end
 TextWidthLimiter(io::IO, limit) = TextWidthLimiter(io, 0, limit)
 has_space(limiter::TextWidthLimiter, width::Int) = limiter.width + width < limiter.limit - 1
 has_space(limiter::TextWidthLimiter, s) = has_space(limiter, textwidth(string(s)))
+has_space(::IO, s) = true
```
aviatesk, line 118: Is this method used ?

test/runtests.jl
```diff
@@ -165,13 +168,16 @@ end
...
+        io = IOBuffer()
+        show(io, first(callsites))
+        @test occursin("< uncached >", String(take!(io)))
         # TODO do some test with `LimitedCallInfo`, but they happen at deeper callsites
     end
 end
 
 @testset "union-split constant-prop'ed callsites" begin
     # constant prop' on all the splits
+    io = IOBuffer()
```
aviatesk, line 180: I'd like to make sure to use different `io` for each test case, in order to eliminate the possibilities of implicit test dependencies (although you're doing `take!(io)` in each test case).

aviatesk (approved):
I like the changes.

src/callsite.jl
timholy: Yes, in tests like
```julia
io = IOBuffer()
Cthulhu.show_callinfo(io, callinfo)
```

test/runtests.jl
timholy: Done

src/callsite.jl
timholy: @simeonschaub, if you still think this needs doing I'm happy to do it in a future PR.

PR #182, opened 2021-07-31
handle pure callsites as special callinfo

aviatesk:
Follows up #180.

@timholy What do you think on this ? I don't think it hurts if we can't recurse into a pure callsite (just because inference doesn't really recurse into it)

### DIFF
--- a/src/callsite.jl
+++ b/src/callsite.jl
@@ -33,10 +33,13 @@ struct UncachedCallInfo <: WrappedCallInfo
     wrapped::CallInfo
 end
 
-# Pure
-struct PureCallInfo <: WrappedCallInfo
-    wrapped::CallInfo
+struct PureCallInfo <: CallInfo
+    argtypes::Vector{Any}
+    rt
+    PureCallInfo(argtypes::Vector{Any}, @nospecialize(rt)) =
+        new(argtypes, rt)
 end
+get_mi(::PureCallInfo) = nothing
 
 # Failed
 struct FailedCallInfo <: CallInfo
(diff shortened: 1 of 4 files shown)


src/reflection.jl
```diff
@@ -51,6 +51,15 @@ function find_callsites(interp::CthulhuInterpreter, CI::Union{Core.CodeInfo, IRC
...
                 rt = argextype(SSAValue(id), CI, sptypes, slottypes)
+                # in unoptimized IR, there may be `slot = rhs` expressions, which `argextype` doesn't handle
+                # so extract rhs for such an case
+                args = c.args
+                if !optimize
+                    if isexpr(c, :(=))
+                        args = c.args[2].args
+                    end
+                end
+                types = mapany(arg -> widenconst(argextype(arg, CI, sptypes, slottypes)), args)
```
timholy, line 62: Very good to move this, I was curious about its former location.

test/runtests.jl
```diff
@@ -63,9 +63,15 @@ isordered(::Type{T}) where {T<:AbstractDict} = false
     @test isempty(callsites)
 
     callsites = find_callsites_by_ftt(iterate, Tuple{SVector{3,Int}, Tuple{SOneTo{3}}}; optimize=false)
-    @test occursin("< pure, uncached >", string(callsites[1]))
+    @test occursin("< pure >", string(callsites[1]))
     @test occursin(r"< constprop > getindex\(::.*Const.*,::.*Const\(1\)\)::.*Const\(1\)", string(callsites[2]))
 
+    # handle pure
```
timholy, line 69: Move this comment up above line 65?

timholy (approved):
This looks great and I approve it to replace #180.

src/callsite.jl
```diff
@@ -207,6 +210,13 @@ function show_callinfo(limiter, ci::Union{MultiCallInfo, FailedCallInfo, Generat
     __show_limited(limiter, name, tt, rt)
 end
 
+function show_callinfo(limiter, (; argtypes, rt)::PureCallInfo)
```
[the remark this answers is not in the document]
aviatesk: Yeah, this syntax fits into my coding style very quickly. Really appreciated, @simeonschaub !

PR #189, opened 2021-08-02
RFC: add terminal tests and add a bit of UI polish

timholy:
This adds some tests of Cthulhu's interactivity. One motivation is simply to improve test coverage, which seems more important as this package gets more widely used. Another motivation is that this is the only way to test the Revise functionality, which was broken for a long time but hopefully fixed in #186.

Finally, I think there are some fixes needed to Cthulhu's interactivity, and having tests makes it easier to improve the code (even if the result of the test changes slightly). As a down payment on that, I've added a couple of blank lines that I think significantly help readability (helping users identify start/stop of new codeviews). Here are some of the other things I've noticed:
- [ ] spurious indentation (sometimes, but not always) in front of "Select a call to descend into". This might actually be left over from Julia's printing. This is currently marked `@test_broken`
- [ ] when you're viewing [S]ource, [A]st, [L]LVM, or [N]ative, it's not obvious how to go back to typed. Should we add a [T]yped to the list?
- [ ] toggling debug state is initially confusing because there are 3 states, not 2, but 2 of the 3 seem quite similar (might require docs)
- [ ] the Variables section seems to be printed based on both the optimization state and the the `iswarn` state, but I'm not certain the latter makes sense
- [ ] it's sometimes hard to know what settings you're in currently; just hitting keys until you get something you like seems to be the strategy I've adopted, but it might be worth asking whether this can be made more rational

I'm not really intending to fix all of those as part of this PR, but we might discuss whether some of these should be addressed.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -3,6 +3,7 @@ module Cthulhu
 using CodeTracking: definition, whereis
 using InteractiveUtils
 using UUIDs
+using REPL: REPL, AbstractTerminal
 
 using Core: MethodInstance
 const Compiler = Core.Compiler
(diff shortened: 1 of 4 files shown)


simeonschaub:
> toggling debug state is initially confusing because there are 3 states, not 2, but 2 of the 3 seem quite similar (might require docs)

The current state is that we support  both inline as well as more verbose debuginfo for optimized IR, but for unoptimized IR, there is only the more verbose one. If we decide inline debuginfo would be useful here, it shouldn't be too hard to add support for that in Base, but I haven't bothered until now since unoptimized IR is usually more concise anyways, so I found the more verbose printing to generally work fine. Where should docs for this go?

Otherwise 100% agree with the points you brought up!

vchuravy:
This is fantastic! Thank you Tim for figuring this out.

timholy:
Great, glad for the positive feedback. Since this doesn't really change much, I'll merge and then start hitting the items in that list in separate PRs. If anyone else wants to help out with this kind of stuff, we do have several CallInfos that we nominally handle but for which there are no tests.

PR #190, opened 2021-08-03
Test UI for call selection, multicall sites, tasks

timholy:
Also add inline cost and ascend


timholy:
Since this is essentially just tests, I'll merge.

timholy:
It's been a good week for code coverage:

xde0noc7

And #191 will increase it further.

Issue #94, opened 2020-10-16
Cthulu does not use limited printing when displaying types.


KristofferC:
Try

```jl
@descend display([0x0001//0x0002])
```

if you have a big REPL history / slow terminal you might have to wait for a while because it tries to display the fill REPL history which is stored inside the repl object.

Now compare with 

```jl
julia> @code_typed display([0x0001//0x0002])
```

There, the type printing of the long vectors is limited.

PR #191, opened 2021-08-03
exhaustive (yet very basic) printer test, code clean up

aviatesk:
- exhaustive printer test
- clean up unnecessary `kwargs`

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -212,7 +212,7 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
     params=current_params(), optimize::Bool=true, interruptexc::Bool=true,
     iswarn::Bool=false, verbose=true, inline_cost::Bool=false)
     if isa(debuginfo, Symbol)
-        debuginfo = getfield(DInfo, selected)::DebugInfo
+        debuginfo = getfield(DInfo, debuginfo)::DebugInfo
     end
 
     is_cached(key) = haskey(optimize ? interp.opt : interp.unopt, key)
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -37,6 +37,11 @@ end
 """
 const CONFIG = CthulhuConfig()
 
+module DInfo
+    @enum DebugInfo none compact source
+end
+using .DInfo: DebugInfo
+
 include("interpreter.jl")
 include("callsite.jl")
 include("reflection.jl")
@@ -197,11 +202,6 @@ function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool)
     (codeinf, rt, infos, slottypes::Vector{Any})
 end
 
-module DInfo
-    @enum DebugInfo none compact source
-end
-using .DInfo: DebugInfo
-
 ##
 # _descend is the main driver function.
 # src/reflection.jl has the tools to discover methods
@@ -257,7 +257,9 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
             ci = preprocess_ci!(codeinf, mi, optimize, CONFIG)
             callsites = find_callsites(interp, codeinf, infos, mi, slottypes, optimize; params, kwargs...)
 
-            display_CI && cthulhu_typed(term.out_stream::IO, debuginfo_key, codeinf, rt, mi, iswarn, verbose, inline_cost)
+            display_CI && cthulhu_typed(term.out_stream::IO, debuginfo_key,
+                codeinf, rt, mi; 
+                iswarn, verbose, inline_cost)
             display_CI = true
         end
 
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -207,11 +207,11 @@ end
 # src/reflection.jl has the tools to discover methods
 # src/ui.jl provides the user facing interface to which _descend responds
 ##
-function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::MethodInstance; override::Union{Nothing, InferenceResult} = nothing, iswarn::Bool, params=current_params(), optimize::Bool=true, interruptexc::Bool=true, verbose=true, inline_cost::Bool=false, kwargs...)
-    debuginfo = DInfo.compact # default is compact debuginfo
-    if :debuginfo in keys(kwargs)
-        selected = kwargs[:debuginfo]
-        # TODO: respect default
+function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::MethodInstance;
+    override::Union{Nothing,InferenceResult}=nothing, debuginfo::Union{Symbol,DebugInfo}=DInfo.compact, # default is compact debuginfo
+    params=current_params(), optimize::Bool=true, interruptexc::Bool=true,
+    iswarn::Bool=false, verbose=true, inline_cost::Bool=false)
+    if isa(debuginfo, Symbol)
         debuginfo = getfield(DInfo, selected)::DebugInfo
     end
 
@@ -220,8 +220,6 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
     menu_options = (cursor = '•', scroll_wrap = true)
     display_CI = true
     while true
-        debuginfo_key = Symbol(debuginfo)
-
         if override === nothing && optimize && interp.opt[mi].inferred === nothing
             # This was inferred to a pure constant - we have no code to show,
             # but make something up that looks plausible.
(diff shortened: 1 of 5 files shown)


src/codeview.jl
```diff
@@ -76,16 +76,22 @@ function is_type_unstable(code::Union{IRCode, CodeInfo}, idx::Int, used::BitSet)
...
 
-cthulhu_warntype(args...) = cthulhu_warntype(stdout::IO, args...)
-function cthulhu_warntype(io::IO, src, rettype, debuginfo, stable_code, inline_cost=false)
+cthulhu_warntype(args...; kwargs...) = cthulhu_warntype(stdout::IO, args...; kwargs...)
+function cthulhu_warntype(io::IO, debuginfo::Union{DebugInfo,Symbol},
+    src::Union{CodeInfo,IRCode}, @nospecialize(rt), mi::Union{Nothing,MethodInstance} = nothing;
+    verbose::Bool=true, inline_cost::Bool=false)
     if inline_cost
-        error("Need a MethodInstance to show inlining costs. Call `cthulhu_typed` directly instead.")
+        @assert isa(mi, MethodInstance) "`mi::MethodInstance` is required"
```
simeonschaub, line 84: This probably shouldn't use `@assert` since IIUC, it is also meant to be called directly and there have been discussions about disabling `@assert` statements by default in the future.

src/codeview.jl
```diff
@@ -76,16 +76,22 @@ function is_type_unstable(code::Union{IRCode, CodeInfo}, idx::Int, used::BitSet)
     return (idx in used) && type isa Type && (!Base.isdispatchelem(type) || type == Core.Box)
 end
 
-cthulhu_warntype(args...) = cthulhu_warntype(stdout::IO, args...)
-function cthulhu_warntype(io::IO, src, rettype, debuginfo, stable_code, inline_cost=false)
+cthulhu_warntype(args...; kwargs...) = cthulhu_warntype(stdout::IO, args...; kwargs...)
+function cthulhu_warntype(io::IO, debuginfo::Union{DebugInfo,Symbol},
+    src::Union{CodeInfo,IRCode}, @nospecialize(rt), mi::Union{Nothing,MethodInstance} = nothing;
```
simeonschaub, line 81:
```suggestion
    src::Union{CodeInfo,IRCode}, @nospecialize(rt), mi::Union{Nothing,MethodInstance}=nothing;
```

src/Cthulhu.jl
```diff
@@ -197,21 +202,16 @@ function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool)
...
-    debuginfo = DInfo.compact # default is compact debuginfo
-    if :debuginfo in keys(kwargs)
-        selected = kwargs[:debuginfo]
-        # TODO: respect default
+function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::MethodInstance;
+    override::Union{Nothing,InferenceResult}=nothing, debuginfo::Union{Symbol,DebugInfo}=DInfo.compact, # default is compact debuginfo
+    params=current_params(), optimize::Bool=true, interruptexc::Bool=true,
+    iswarn::Bool=false, verbose=true, inline_cost::Bool=false)
+    if isa(debuginfo, Symbol)
         debuginfo = getfield(DInfo, selected)::DebugInfo
```
timholy, line 215: Where does `selected` get defined now?

src/Cthulhu.jl
```diff
@@ -197,21 +202,16 @@ function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool)
...
 ##
-function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::MethodInstance; override::Union{Nothing, InferenceResult} = nothing, iswarn::Bool, params=current_params(), optimize::Bool=true, interruptexc::Bool=true, verbose=true, inline_cost::Bool=false, kwargs...)
-    debuginfo = DInfo.compact # default is compact debuginfo
-    if :debuginfo in keys(kwargs)
-        selected = kwargs[:debuginfo]
-        # TODO: respect default
+function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::MethodInstance;
+    override::Union{Nothing,InferenceResult}=nothing, debuginfo::Union{Symbol,DebugInfo}=DInfo.compact, # default is compact debuginfo
+    params=current_params(), optimize::Bool=true, interruptexc::Bool=true,
+    iswarn::Bool=false, verbose=true, inline_cost::Bool=false)
```
simeonschaub, line 213: Is `verbose` the best name here? It feels more like not hiding code from you. Perhaps `hide_type_stable` would be clearer?

src/codeview.jl
```diff
@@ -129,7 +137,7 @@ function cthulhu_typed(io::IO, debuginfo, src, rt, mi, iswarn, stable_code, inli
     end
     postprinter = iswarn ? InteractiveUtils.warntype_type_printer : IRShow.default_expr_type_printer
 
-    should_print_stmt = (iswarn || src isa IRCode || stable_code) ? Returns(true) : is_type_unstable
+    should_print_stmt = (iswarn || src isa IRCode || verbose) ? Returns(true) : is_type_unstable
```
timholy, line 140: This is not anything this PR needs to address, but I just added a comment about `verbose` in #189. Do you agree that the sense seems flipped?

timholy (commented):
A very nice cleanup, thanks!

How hard do you think it would be to add more than "it works" tests? We may be covered elsewhere, of course, but the terminal tests I've added in #190 do get to some of that. So I don't think it's urgent here, and having exhaustive coverage has real advantages.

simeonschaub (approved):
Thanks for doing this, I really like the added consistency and clarity! This is technically a breaking user-facing change though, are we fine with that for a minor version? Would be good to at least get another opinion on this.

Awesome to finally have some basic tests! I have put this off for a while, but it's great to at least be certain that these don't start throwing errors.

timholy:
> This is technically a breaking user-facing change though

Which part? I missed that bit.

Agreed, though, that the `verbose` option is pretty confusingly-named. Changing that would, of course, be a breaking change, but we could make it a deprecation instead.

simeonschaub:
> > This is technically a breaking user-facing change though
> 
> Which part? I missed that bit.

It changes the order of arguments to `cthulhu_code_typed`/`cthulhu_code_warntype` and makes some of the arguments kwargs instead.

timholy:
Doesn't that only affect the non-exported `cthulhu_typed`/`cthulhu_warntype`? I don't see any changes to the `descend_*` API.

timholy:
Once we get this merged, I have a fix for #94 ready to go (it will likely conflict with this PR so I'll wait until this merges), and then I think it's time for a new release. Addressing the UI issues in #189 really should be a minor bump, not a patch bump.

src/Cthulhu.jl
aviatesk: Good catch, it's just because I didn't run JET on this PR :P

src/codeview.jl
aviatesk: Thanks ! Now I just remember I got the same pointer from you on Julia Base 😅 Will make sure to use explicit error.

aviatesk:
> How hard do you think it would be to add more than "it works" tests?

I'd like to add more specific and individual tests in the future. For example, these test cases don't cover any cases when `src::CodeInfo`, while `cthulhu_typed` really changes its behavior based on the type of `src`.
I think we need to setup each specific test cases for them, and I'd like to gradually enrich our test suite.

aviatesk:
> Addressing the UI issues in #189 really should be a minor bump, not a patch bump.

Yeah, I'm just fine with a minor bump.

aviatesk:
> Is verbose the best name here? It feels more like not hiding code from you. Perhaps hide_type_stable would be clearer?
> Agreed, though, that the verbose option is pretty confusingly-named. Changing that would, of course, be a breaking change, but we could make it a deprecation instead.

yeah, we should rename this name. Can we just address this in a separate PR though ? Just for the sake of easier reviews.

timholy:
Right, assuming I'm not wrong about https://github.com/JuliaDebug/Cthulhu.jl/pull/191#issuecomment-891761616, our next release can be a patch release. Then fixing some of the issues in #189 would be a minor release since it will have significant user-visible change.

simeonschaub:
> Doesn't that only affect the non-exported `cthulhu_typed`/`cthulhu_warntype`? I don't see any changes to the `descend_*` API.

Yeah, I thought they were part of the public API as well, but this is probably fine then with the minor version bump.

>  I think we need to setup each specific test cases for them, and I'd like to gradually enrich our test suite.

Yeah, we could probably just write the outputs of different configurations to a file and always test against that. The only worry would be to keep that in sync with Base if the printing there changes slightly or the optimizer gets more aggressive, but that's probably doable.

timholy:
I don't think any of the [changes](https://github.com/JuliaDebug/Cthulhu.jl/compare/v2.1.2...master) need a minor version bump, as they all seem to be bugfixes, testing, and cleanup (which is patch territory). But either is OK with me, minor bumps are generally not disruptive post-1.0.

PR #193, opened 2021-08-03
change "verbose" terminology to "hide"

aviatesk:
As discussed in #189 and #191, the sense of "verbose" option seems to be
flipped, and its behavior is a bit confusing.

In this PR, I'd propose:
- change "verbose" to "hide" (type-stable statements)
- make it `false` by default
- respect this configuration whenever if we're inspecting `IRCode` or `CodeInfo`,
  or in warntype mode

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -212,7 +212,7 @@ end
 function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::MethodInstance;
     override::Union{Nothing,InferenceResult}=nothing, debuginfo::Union{Symbol,DebugInfo}=DInfo.compact, # default is compact debuginfo
     params=current_params(), optimize::Bool=true, interruptexc::Bool=true,
-    iswarn::Bool=false, hide::Bool=false, inline_cost::Bool=false)
+    iswarn::Bool=false, hide_type_stable::Bool=false, inline_cost::Bool=false)
     if isa(debuginfo, Symbol)
         debuginfo = getfield(DInfo, debuginfo)::DebugInfo
     end
@@ -259,7 +259,7 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
 
             display_CI && cthulhu_typed(term.out_stream::IO, debuginfo,
                 codeinf, rt, mi;
-                iswarn, hide, inline_cost)
+                iswarn, hide_type_stable, inline_cost)
             display_CI = true
         end
 
@@ -304,7 +304,7 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
                 _descend(term, next_interp, next_mi;
                          debuginfo,
                          params, optimize, interruptexc,
-                         iswarn, hide, inline_cost)
+                         iswarn, hide_type_stable, inline_cost)
                 continue
             end
 
@@ -321,12 +321,12 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
             _descend(term, interp, next_mi;
                      override = isa(info, ConstPropCallInfo) ? info.result : nothing, debuginfo,
                      params,optimize, interruptexc,
-                     iswarn, hide, inline_cost)
+                     iswarn, hide_type_stable, inline_cost)
 
         elseif toggle === :warn
             iswarn ⊻= true
-        elseif toggle === :hide
-            hide ⊻= true
+        elseif toggle === :hide_type_stable
+            hide_type_stable ⊻= true
         elseif toggle === :optimize
             optimize ⊻= true
             if !is_cached(mi)
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -213,9 +213,6 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
     override::Union{Nothing,InferenceResult}=nothing, debuginfo::Union{Symbol,DebugInfo}=DInfo.compact, # default is compact debuginfo
     params=current_params(), optimize::Bool=true, interruptexc::Bool=true,
     iswarn::Bool=false, hide_type_stable::Union{Nothing,Bool}=nothing, verbose::Union{Nothing,Bool}=nothing, inline_cost::Bool=false)
-    if !(isnothing(verbose) || isnothing(hide_type_stable))
-        error("`verbose` is being replaced with `hide_type_stable`, do not use both")
-    end
     if isnothing(hide_type_stable)
         hide_type_stable = something(verbose, false)
     end
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -219,7 +219,7 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
     if isnothing(hide_type_stable)
         hide_type_stable = something(verbose, false)
     end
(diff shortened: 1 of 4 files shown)


simeonschaub:
I am usually a fan of longer, more descriptive names, what are your thoughts on naming it `hide_type_stable` instead? `hide` seems potentially ambiguous to me.

aviatesk:
yep, I'm willing to change to that name. Thanks for your suggestion !

src/Cthulhu.jl
```diff
@@ -212,7 +212,7 @@ end
 function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::MethodInstance;
     override::Union{Nothing,InferenceResult}=nothing, debuginfo::Union{Symbol,DebugInfo}=DInfo.compact, # default is compact debuginfo
     params=current_params(), optimize::Bool=true, interruptexc::Bool=true,
-    iswarn::Bool=false, verbose=true, inline_cost::Bool=false)
+    iswarn::Bool=false, hide_type_stable::Bool=false, inline_cost::Bool=false)
```
timholy, line 215: If we want to make this non-erroring for anyone who supplies `verbose` directly, then perhaps use
```julia
verbose=nothing, hide_type_stable=nothing, ...
```
and then later
```julia
if !(verbose === nothing || hide_type_stable === nothing)
    error("`verbose` is being replaced with `hide_type_stable`, do not use both")
end
if hide_type_stable === nothing
    hide_type_stable = something(verbose, false)
end
```
although even better would be to `depwarn` the use of `verbose`.

I'm not insisting on this; UI changes are kind of a gray zone as far as I am concerned. If others think this is more hassle than desirable, I'm fine with the way you've done it.

timholy (commented):
Let's make this part of 2.2

src/Cthulhu.jl
aviatesk: Yeah, that should be better.

aviatesk: We also want to keep supporting <kbd>v</kbd> ... ? I'd like to avoid it, since I guess it will complicate the UI.

src/Cthulhu.jl
```diff
@@ -212,7 +212,14 @@ end
 function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::MethodInstance;
     override::Union{Nothing,InferenceResult}=nothing, debuginfo::Union{Symbol,DebugInfo}=DInfo.compact, # default is compact debuginfo
     params=current_params(), optimize::Bool=true, interruptexc::Bool=true,
-    iswarn::Bool=false, verbose=true, inline_cost::Bool=false)
+    iswarn::Bool=false, hide_type_stable::Union{Nothing,Bool}=nothing, verbose::Union{Nothing,Bool}=nothing, inline_cost::Bool=false)
+    if !(isnothing(verbose) || isnothing(hide_type_stable))
+        error("`verbose` is being replaced with `hide_type_stable`, do not use both")
+    end
```
simeonschaub, line 218: We probably don't really need to make this an error if we have the deprecation warning.

src/Cthulhu.jl
```diff
@@ -212,7 +212,14 @@ end
...
+    if isnothing(hide_type_stable)
+        hide_type_stable = something(verbose, false)
+    end
+    isnothing(verbose) || @warn "use `hide_type_stable` instead of `verbose`." maxlog=1
```
simeonschaub, line 222:
```suggestion
    isnothing(verbose) || Base.depwarn("The `verbose` keyword argument to `Cthulhu.descend` is deprecated. Use `hide_type_stable` instead.")
```

src/Cthulhu.jl
simeonschaub: Yes, dropping support for that should be fine.

simeonschaub (approved):
Some minor comments, but otherwise looks good!

src/Cthulhu.jl
timholy: If the user gives both and they conflict, which one wins?

simeonschaub: The non-deprecated one. But it doesn't really mater, we might as well leave this as-is.

test/runtests.jl
```diff
@@ -677,19 +677,87 @@ Revise.track(CthulhuTestSandbox, normpath(@__DIR__, "sandbox.jl"))
...
+
+        let # by default, should print every statement
+            s = prints()
+            @test occursin("globalvar", s)
+            @test occursin("undefvar", s)
+        end
+        let # should omit type stable statements
+            s = prints(; hide_type_stable=true)
+            @test !occursin("globalvar", s)
+            @test_broken occursin("undefvar", s)
```
aviatesk, line 744: I found the current implementation of `is_type_unstable` may not work well for unoptimized code.

Especially, it doesn't handle `lhs = rhs` statements where `rhs` is type unstable ?
```julia
julia> const globalvar = Ref(42)
Base.RefValue{Int64}(42)

julia> descend(; optimize=false) do
           a = sin(globalvar[])
           b = sin(undefvar)
           return (a, b)
       end
│ ─ %-1  = invoke #3()::Tuple{Float64, Any}
    @ REPL[4]:2 within `#3`
1 ─ %1 = Base.getindex(Main.globalvar)::Int64
│        (a = Main.sin(%1))::Float64
│   @ REPL[4]:3 within `#3`
│        (b = Main.sin(Main.undefvar))::Any
│   @ REPL[4]:4 within `#3`
│   %4 = Core.tuple(a, b)::Tuple{Float64, Any}
└──      return %4
Toggles: [o]ptimize, [w]arn, [h]ide type-stable statements, [d]ebuginfo, [i]nlining costs, [s]yntax highlight for Source/LLVM/Native.
 • %1  = getindex(::Base.RefValue{Int64})::Int64
   %2  = sin(::Int64)::Float64
   ↩

│ ─ %-1  = invoke #3()::Tuple{Float64, Any}
│   @ REPL[4]:4 within `#3`
│   %4 = Core.tuple(a, b)::Tuple{Float64, Any}
Toggles: [o]ptimize, [w]arn, [h]ide type-stable statements, [d]ebuginfo, [i]nlining costs, [s]yntax highlight for Source/LLVM/Native.
 • %1  = getindex(::Base.RefValue{Int64})::Int64
   %2  = sin(::Int64)::Float64
   ↩
```

simeonschaub: Hmm, yeah, we might need to special-case assignments to slots.

aviatesk: I found it `idx in used` returns `false` for unoptimized code:
https://github.com/JuliaDebug/Cthulhu.jl/blob/93743cb4efe0ddfd189808ec589449ef97361387/src/codeview.jl#L76

So maybe we need to add support for unoptimized code case within Julia Base ?

simeonschaub: Hmm, I thought that should already work 🤔. I can look into it later today. Probably doesn't need to block this PR though.

simeonschaub:
> Let's make this part of 2.2

Oops, didn't realize we never released this! It's now part of 2.1.4, should we just leave it like that now? Or should we yank it?

aviatesk:
It should be better to ship this in v2.2, but I guess it's okay to leave it for now, since we still support `verbose` option (although its meaning is flipped now) and it shouldn't break user code in a "serious" way.

simeonschaub:
Should we adopt a policy of always tagging a new release for any user-facing changes? That has worked quite well for ChainRules.

aviatesk:
yeah, I don't have a strong opinion on release management, but as far as we can use manor version bump to take in "small" user-facing changes, I'm just fine with it.

PR #194, opened 2021-08-03
JET on Cthulhu

aviatesk:
Ran JET on Cthulhu, and fixed some real bugs and dead code, and improved type stabilities:
```julia
julia> using JET; report_package(Cthulhu)
```

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -266,7 +266,7 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
             else
                 (codeinf, rt, infos, slottypes) = lookup(interp, mi, optimize)
             end
-            codeinf = preprocess_ci!(codeinf, mi, optimize, CONFIG)
+            ci = preprocess_ci!(codeinf, mi, optimize, CONFIG)
             callsites = find_callsites(interp, codeinf, infos, mi, slottypes, optimize; params)
 
             display_CI && cthulhu_typed(term.out_stream::IO, debuginfo,
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -175,7 +175,8 @@ function codeinst_rt(code::CodeInstance)
     end
 end
 
-function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool)
+# `Base.@aggressive_constprop` here in order to make sure the constant propagation of `allow_no_codeinf`
+Base.@aggressive_constprop function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool; allow_no_codeinf::Bool=false)
     if !optimize
         codeinf = copy(interp.unopt[mi].src)
         rt = interp.unopt[mi].rt
@@ -193,12 +194,19 @@ function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool)
             codeinf = ir
             infos = ir.stmts.info
             slottypes = ir.argtypes
-        else
+        elseif allow_no_codeinf
             # This doesn't showed up as covered, but it is (see the CI test with `coverage=false`).
             # But with coverage on, the empty function body isn't empty due to :code_coverage_effect expressions.
             codeinf = nothing
             infos = []
             slottypes = Any[Base.unwrap_unionall(mi.specTypes).parameters...]
+        else
+            Core.eval(Main, quote
+                interp = $interp
+                mi = $mi
+                optimize = $optimize
+            end)
+            throw("couldn't find the source; inspect `Main.interp` and `Main.mi`")
         end
     end
     (codeinf, rt, infos, slottypes::Vector{Any})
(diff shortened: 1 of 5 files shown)


src/backedges.jl
```diff
@@ -105,17 +104,17 @@ function callstring(io, sfs::Vector{StackTraces.StackFrame})
...
     return callstring(io, instance(sfs)) * string(" at ", sf.file, ':', sf.line)
 end
-callstring(io, ipframes::Vector{IPFrames}) = isempty(ipframes) ? "" : callstring(io, ipframes[1].sfs)
+callstring(io::IO, ipframes::Vector{IPFrames}) = isempty(ipframes) ? "" : callstring(io, ipframes[1].sfs)
 
-struct Data{T}
+struct Data
     callstr::String
-    nd::T
+    nd::MethodInstance
```
timholy, line 111: Can't do this, it will break SnoopCompile. I think that's true of most of the changes in this file. I'll see about submitting a "stub" version of what's in SnoopCompile to protect this properly (sorry, it was bad of me not to do that from the outset).

timholy (commented):
OK, we might need to hold back on this a little bit, see my concerns about SnoopCompile. But I love it and most of it looks great!

src/backedges.jl
aviatesk: ops, thanks. I will revert this part then.

timholy:
Wish-list for JET: create a test that would have caught the bug. That's not asking too much, is it? :slightly_smiling_face:

src/Cthulhu.jl
```diff
@@ -193,12 +194,19 @@ function lookup(interp::CthulhuInterpreter, mi::MethodInstance, optimize::Bool)
...
             codeinf = nothing
             infos = []
             slottypes = Any[Base.unwrap_unionall(mi.specTypes).parameters...]
+        else
+            Core.eval(Main, quote
+                interp = $interp
+                mi = $mi
+                optimize = $optimize
+            end)
+            throw("couldn't find the source; inspect `Main.interp` and `Main.mi`")
```
simeonschaub, line 209: Did you mean to use `error` instead of `throw` here?

src/backedges.jl
simeonschaub: We could also just add some integration tests like e.g. in https://github.com/JuliaDiff/ChainRules.jl/blob/master/.github/workflows/IntegrationTest.yml

vchuravy: I am all for integration tests!

aviatesk: The integration test is nice, thanks so much @simeonschaub !

aviatesk:
> Wish-list for JET: create a test that would have caught the bug. That's not asking too much, is it? 🙂

Yep, I will work on it ;)

src/Cthulhu.jl
```diff
@@ -258,7 +266,7 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
             else
                 (codeinf, rt, infos, slottypes) = lookup(interp, mi, optimize)
             end
-            ci = preprocess_ci!(codeinf, mi, optimize, CONFIG)
+            codeinf = preprocess_ci!(codeinf, mi, optimize, CONFIG)
```
timholy, line 269: Maybe, but see #196. You might test `@descend tryparse(VersionNumber, "v2.1.3")` on this branch.

timholy (approved):
This looks great!

src/Cthulhu.jl
aviatesk: Thanks, I think I have a fix for #196

aviatesk: I will address this issue in a separate PR

src/backedges.jl
```diff
@@ -86,17 +86,17 @@ specTypes(mi::MethodInstance) = mi.specTypes
...
 
-instance(sfs::Vector{StackTraces.StackFrame}) = isempty(sfs) ? Core.Compiler.Timings.ROOTmi : sfs[end].linfo
+instance(sfs::Vector{StackTraces.StackFrame}) = isempty(sfs) ? Core.Compiler.Timings.ROOTmi : sfs[end].linfo::MethodInstance # we checked this type condition within `buildframes`
 method(sfs::Vector{StackTraces.StackFrame}) = method(instance(sfs))
 
 instance(ipframes::Vector{IPFrames}) = isempty(ipframes) ? Core.Compiler.Timings.ROOTmi : instance(ipframes[1].sfs)
 backedges(ipframes::Vector{IPFrames}) = (ret = ipframes[2:end]; isempty(ret) ? () : (ret,))
 
-function callstring(io, mi)
+function callstring(io::IO, mi::MethodInstance)
```
timholy, line 95: Oops, this one was bad too (broke SnoopCompile). I will fix.

aviatesk: Agh, sorry I just checked the result of the integration test.

timholy: The mistake is mine. I just assumed we'd be fine, but I forgot that until #189 I didn't know how to test the interactive menus, and so SnoopCompile evidently doesn't have complete integration tests. I'll try to fix that, but meanwhile no hurry.

PR #197, opened 2021-08-04
A friendlier face for the squid

timholy:
This adds several elements of UI polish first identified in #189:
- eliminates spurious indentation
- adds a [T]yped code-selector
- encodes state by the color of toggles & selectors
- prints Variables for unoptimized typed code regardless of iswarn

It also prints the method being inspected in bold at the top of the code.

The best way to review is probably just to try it :smile:.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -270,15 +270,16 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
                         cthulhu_typed(io, debuginfo, codeinf, rt, mi;
                                     iswarn, hide_type_stable, inline_cost)
                     end
-                    rmatch = match(r"\u001B\[90m\u001B\[(\d+)G( *)\u001B\[1G\u001B\[39m\u001B\[90m( *)\u001B\[39m$", str)
+                    rmatch = findfirst(r"\u001B\[90m\u001B\[(\d+)G( *)\u001B\[1G\u001B\[39m\u001B\[90m( *)\u001B\[39m$", str)
                     if rmatch !== nothing
-                        str = chop(str, tail = 27 + sum(length, rmatch.captures))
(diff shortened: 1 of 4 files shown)


simeonschaub:
This is awesome! Super cool!

One small issue I noticed was that if you are looking at anything that's not typed IR and select one of the toggles above, it will go back to showing the typed IR, but it won't update the menu below to show that it's now displaying typed IR.

src/Cthulhu.jl
```diff
@@ -261,14 +262,30 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
...
+                printstyled(IOContext(term.out_stream::IO, :limit=>true), mi.def, '\n'; bold=true)
+                if debuginfo == DInfo.compact
+                    # Eliminate trailing indentation (see first item in bullet list in PR #189)
+                    str = stringify() do io
+                        cthulhu_typed(io, debuginfo, codeinf, rt, mi;
+                                    iswarn, hide_type_stable, inline_cost)
+                    end
+                    rmatch = match(r"\u001B\[90m\u001B\[(\d+)G( *)\u001B\[1G\u001B\[39m\u001B\[90m( *)\u001B\[39m$", str)
+                    if rmatch !== nothing
+                        str = chop(str, tail = 27 + sum(length, rmatch.captures))
```
vchuravy, line 275: What is the magic number 27?

vchuravy (approved):

src/Cthulhu.jl
timholy: It's the number of characters in that match string aside from what's in the captures. I'll add a comment.

aviatesk (approved):
nice enhancements !

src/Cthulhu.jl
timholy: Or better yet, I'll just use `rmatch.offset`.

PR #198, opened 2021-08-04
fix alignment of inlining cost with verbose dinfo

simeonschaub:
Before:
image

After:
image


vchuravy:
Can you add a test as well if that makes sense?

simeonschaub:
I tried out a more systematic approach to testing in #199, since I otherwise couldn't come up with a good way of testing something like this. Please let me know what you think.

vchuravy:
Great thanks!

Issue #196, opened 2021-08-03
Inconsistency in handling the return of preprocess_ci!


timholy:
https://github.com/JuliaDebug/Cthulhu.jl/blob/9fdc45c233cf593bc426fb6aea6f16095efe001d/src/Cthulhu.jl#L261-L262

returns `ci` but the later code uses `codeinf`. Is this deliberate? Note it's different from

https://github.com/JuliaDebug/Cthulhu.jl/blob/9fdc45c233cf593bc426fb6aea6f16095efe001d/test/runtests.jl#L28-L42

where the return value is used henceforth. The latter is similar to #177, which fixed some problems, and that leads me to think that the one in `src/Cthulhu.jl` must be wrong. But in fact if you use the one in `test/runtests.jl`, here's what you get:

```julia
ulia> @find_callsites_by_ftt tryparse(VersionNumber, "v2.1.3")
ERROR: AssertionError: argextype only works on argument-position values
Stacktrace:
 [1] argextype(x::Any, src::Core.Compiler.IRCode, sptypes::Vector{Any}, slottypes::Vector{Any})
   @ Core.Compiler ./compiler/utilities.jl:229
 [2] (::Cthulhu.var"#13#16"{Core.Compiler.IRCode, Vector{Any}, Vector{Any}})(arg::Expr)
   @ Cthulhu ~/.julia/dev/Cthulhu/src/reflection.jl:46
 [3] map!(f::Cthulhu.var"#13#16"{Core.Compiler.IRCode, Vector{Any}, Vector{Any}}, dest::Vector{Any}, A::Vector{Any})
   @ Base ./abstractarray.jl:2794
 [4] mapany(f::Function, A::Vector{Any})
   @ Base ./abstractarray.jl:2803
 [5] find_callsites(interp::Cthulhu.CthulhuInterpreter, CI::Core.Compiler.IRCode, stmt_info::Vector{Any}, mi::Core.MethodInstance, slottypes::Vector{Any}, optimize::Bool; params::Core.Compiler.NativeInterpreter)
   @ Cthulhu ~/.julia/dev/Cthulhu/src/reflection.jl:46
 [6] find_callsites
   @ ~/.julia/dev/Cthulhu/src/reflection.jl:28 [inlined]
 [7] find_callsites_by_ftt(f::Any, TT::Any; optimize::Bool)
   @ Main ~/.julia/dev/Cthulhu/test/runtests.jl:40
[...]

PR #199, opened 2021-08-04
test all IRShow combinations

simeonschaub:
This tests all IRShow combinations against hardcoded test files. Is this
reasonable? Or are we worried that this will be a PITA to keep
up-to-date with Base?


timholy:
Interesting idea. I'm pretty attracted by this but I do fear it could be hard to stay up-to-date. But honestly I don't know. Have you tried to see what might have changed between the very first commit in 1.7 and current master?

simeonschaub:
> Have you tried to see what might have changed between the very first commit in 1.7 and current master?

It works fine with the 1.7 beta at least, the first commit likely won't work because there have been some fixes to IR printing in Base since then. Master works just fine as well. It likely wouldn't be hard to make this version-dependent once this starts changing in Base, but I definitely understand your worries.

@tkf Any ideas about the test failure here? Is it a problem to run `dce!` again here?

tkf:
What was the problem with DCE? It seems already solved?

BTW, we can also set up a github action to continuously update this, although it'd still mean someone has to click the merge button from time to time.

Project.toml
```diff
@@ -17,10 +17,12 @@ FoldingTrees = "1"
 julia = "1.7"
 
 [extras]
+DeepDiffs = "ab62b9b5-e342-54a8-a765-a90f495de1a6"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+TestSetExtensions = "98d24dd4-01ad-11ea-1b02-c9a08f80db04"
```
tkf, line 25:
```suggestion
```

Not used anymore?

timholy:
xref #196. I haven't had time to dig into it yet.

simeonschaub:
> What was the problem with DCE? It seems already solved?

I meant the failure [here](https://github.com/JuliaDebug/Cthulhu.jl/runs/3243919090). I now worked arround the issue by only running DCE for the tests where I need to get rid of these expressions. It could definitely be related to the issue Tim referenced.

I now get a segfault on MacOS though, could that be a Julia bug? I noticed we are usually running DCE twice, could that be an issue where only running it once can generate bad IR?

> BTW, we can also set up a github action to continuously update this, although it'd still mean someone has to click the merge button from time to time.

I thought about that as well. The only issue is that we probably want to start running tests on multiple Julia versions again once 1.7 is released, so we might have to make these tests version-dependent eventually. That might be solvable with some clever scripting though.

vchuravy:
Cthulhu is the only real user of the DCE code in Base so yes we could get into accidentally creating illegal code. If you can nail it down to a MWE I would be thankful

simeonschaub:
> Cthulhu is the only real user of the DCE code in Base so yes we could get into accidentally creating illegal code. If you can nail it down to a MWE I would be thankful

It looks like the segfaults are non-deterministic, so tests seem to pass now. We should definitely keep an eye on that though.

test/runtests.jl
```diff
@@ -650,113 +650,12 @@ let callsites = find_callsites_by_ftt(issue152_another, (Tuple{Float64,Vararg{Fl
     @test !isempty(callsites)
 end
 
-# NOTE setup to for `cthulhu_ast`
+# NOTE setup too for `cthulhu_ast`
 include("sandbox.jl")
 using .CthulhuTestSandbox
 Revise.track(CthulhuTestSandbox, normpath(@__DIR__, "sandbox.jl"))
```
aviatesk, line 657: Let's move these lines into test/codeview.jl as well, they're only used by the printer test anyway.

simeonschaub: Done.

timholy:
@vchuravy,

> If you can nail it down to a MWE I would be thankful

There's one in #199: `@find_callsites_by_ftt tryparse(VersionNumber, "v2.1.3")`

vchuravy:
Will push this on my to-do whence I am back from vacation.

aviatesk:
Test outputs need to be updated to take in the changes from #197 , where variable types are rendered in unoptimized code view.

aviatesk (commented):
Not too sure about the development cost involved with updating these test outputs, but I believe it's worth a try.

aviatesk (approved):

simeonschaub:
Yes, if it turns out to be too annoying we can just disable them again.

PR #200, opened 2021-08-04
add integration tests for SnoopCompile

simeonschaub:


timholy (approved):
Thanks for making it so easy!

PR #204, opened 2021-08-05
fix #196

### LINKED ISSUE #196: Inconsistency in handling the return of preprocess_ci!

aviatesk:
Fixes #196 by making sure Cthulhu doesn't call `argextype` on
expressions other than `:call` and `:invoke`, which can result
in error. Other expressions like `:foreigncall` should be handled
by special casing.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -267,7 +267,7 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
             else
                 (codeinf, rt, infos, slottypes) = lookup(interp, mi, optimize)
             end
-            ci = preprocess_ci!(codeinf, mi, optimize, CONFIG)
+            codeinf = preprocess_ci!(codeinf, mi, optimize, CONFIG)
             callsites = find_callsites(interp, codeinf, infos, mi, slottypes, optimize; params)
 
             if display_CI
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -30,7 +30,7 @@ function find_callsites(interp::CthulhuInterpreter, CI::Union{Core.CodeInfo, IRC
     for (id, c) in enumerate(isa(CI, IRCode) ? CI.stmts.inst : CI.code)
         callsite = nothing
         isa(c, Expr) || continue
-        if stmt_info !== nothing
+        if stmt_info !== nothing && is_call_expr(c, optimize)
             info = stmt_info[id]
             if info !== nothing
                 rt = argextype(SSAValue(id), CI, sptypes, slottypes)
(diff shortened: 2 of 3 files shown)


aviatesk:
I confirmed this PR pass tests added in #199 , except the parts I mentioned in https://github.com/JuliaDebug/Cthulhu.jl/pull/199#issuecomment-893426201

timholy (approved):
Woot! Thanks so much!

PR #205, opened 2021-08-05
restore original intent of IRShow test case

simeonschaub:
I think this makes this a better test case, since the small union did
not show up here before.


aviatesk:
We actually need to further update test outputs

vchuravy:
@simeonschaub  bump? Probably needs a rebase as well.

simeonschaub:
The issue here is that my previous approach that replaced `code_coverage_effect` statements with `nothing` and then runs dce on the IR deletes other code here as well, leading to different results when code-coverage is enabled. Is there a better way to delete statements from `IRCode`?

simeonschaub:
Ok, this now uses PerformanceTestTools.jl for the tests instead of manually messing with the IR, which should be a lot more reliable than the current approach. I also moved the organization of the tests around a bit for this, I hope that's ok.

simeonschaub:
> We actually need to further update test outputs ?

Just to clarify, I just made a typo previously, so the tests weren't testing the printing for small unions as I thought. We hopefully don't have to touch these tests after this for a while.

PR #206, opened 2021-08-05
Fix SnoopCompile breakage

timholy:
I should have added a test for this long ago, sorry @aviatesk.


aviatesk (approved):

aviatesk:
Thanks for the fix !

PR #207, opened 2021-08-05
Improve inferrability

timholy:
SnoopCompile detected many new inference problems, most notably a
large number of `Core.Box` slots. Fixing this is the reason for some
substantial code-moves. (Inner functions sometimes seem surprisingly
difficult to fix, so I frequently just move them out and pass arguments.)

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -229,7 +229,7 @@ function _descend(term::AbstractTerminal, interp::CthulhuInterpreter, mi::Method
         debuginfo = getfield(DInfo, debuginfo)::DebugInfo
     end
 
-    is_cached(key) = haskey(optimize ? interp.opt : interp.unopt, key)
+    is_cached(key, opt::Bool) = haskey(opt ? interp.opt : interp.unopt, key)
 
     menu_options = (cursor = '•', scroll_wrap = true)
     display_CI = true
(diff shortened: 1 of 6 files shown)


aviatesk (approved):
Looks solid, but GitHub actions seems extremely slow today ?

timholy:
Yeah, clearly there is something wrong. It's odd that it freezes after the build step (though this apparently succeeds). I tried updating to the latest nightly and I still can't reproduce this locally.

aviatesk:
I confirmed this PR passes the test suite locally on my machine (macOS), so please go ahead to merge if you want.

timholy:
Well, I don't want CI to be borked forever. Let me push some debug commits.

timholy:
In #208 it looks like I identified that the problematic change was `stdout::Base.TTY`, which sort of makes sense as the kind of change that could cause problems for CI.

Some stats: with the following test:
```julia
tinf = @snoopi_deep descend(gcd, (Int, Int))
```
and then hitting 'o' and 'w' followed by 'q', we had:
- master: 108 inference failures
- this PR: 74 inference failures

Almost all remaining inference problems are from printing. `cthulhu_typed` is a pretty egregious offender; to fix that we'd have to stop assigning functions to local variables, and instead split each call, and it still wouldn't come close to eliminating all inference triggers. If we decide to do that, I think we can save that for a future PR.

Issue #209, opened 2021-08-06
Why is Cthulhu slow on long method bodies?


timholy:
To answer my own question...it's almost entirely due to `make_typealias` in `base/show.jl`. https://github.com/JuliaLang/julia/pull/36107. I think there was discussion of caching its results at some point, but I can't find the issue.

KristofferC:
I have one issue about it here https://github.com/JuliaLang/julia/issues/37990 (although it is perhaps badly named)

PR #210, opened 2021-08-06
Reduce latency

timholy:
```
julia> using Cthulhu

julia> tstart = time(); descend(gcd, (Int, Int)); time() - tstart
```
and hit 'q' while you are waiting.

Here's the timing:
- master: 6.57s
- this branch with just precompilation: 5.86s
- this branch with compile=min optimize=1: 4.68s

I haven't run this extensively, so I can't promise that there isn't a hit to runtime performance, but it looks from my analysis in #209 that this module is not performance-sensitive. It's a bit at odds with a couple of annotations we have https://github.com/JuliaDebug/Cthulhu.jl/blob/60d8b5476ad79d6796ce5295c10c312d9e32d0bb/src/Cthulhu.jl#L179
so some discussion may be merited.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -1,5 +1,7 @@
 module Cthulhu
 
+Base.Experimental.@compiler_options compile=min optimize=1
+
 using CodeTracking: definition, whereis
 using InteractiveUtils
 using UUIDs
@@ -524,5 +526,14 @@ Follow a chain of calls (either through a backtrace `bt` or the backedges of a `
 with the option to `descend` into intermediate calls. `kwargs` are passed to [`descend`](@ref).
 """
 ascend
+
+if ccall(:jl_generating_output, Cint, ()) == 1
+    input = Pipe()
+    Base.link_pipe!(input, reader_supports_async=true, writer_supports_async=true)
+    term = REPL.Terminals.TTYTerminal("dumb", input.out, IOBuffer(), IOBuffer())
+    write(input.in, 'q')
+    descend(gcd, (Int, Int); terminal=term)
+    nothing
+end
 
 end


aviatesk (approved):
Looks great.

> It's a bit at odds with a couple of annotations we have

The only purpose of `Base.@aggressive_constant_prop` is to narrow down the return type of `lookup` from `Tuple{Union{Nothing,CodeInfo,IRCode},...}` to `Tuple{Union{CodeInfo,IRCode}}` when not in test, by making sure inference propagates constant information of `allow_no_codeinf` argument.

It may come with inference cost (but we can't say anything solid unless we do benchmarking, because inference performance isn't really monotonic w.r.t. inference accuracy), but since we're precompiling the inference cache in this PR, it won't hurt the latency ?

timholy:
That's correct, because this package "owns" `lookup` so all instances will be cached.

I don't know what's up with the coverage report; it appeared shortly after submission before all workers had finished, and hasn't updated. I don't think it's a permanent state of affairs, but feel free to close and re-open if you want to run it again.

aviatesk:
Hmmm, very weird, the coverage is still very low. 
Well, I generally don't care codecov report so seriously, so I'd like to merge this PR as is. The latency gain is really valuable imho.

timholy:
Sure, we can fix the report later (if the breakage is real, which it may not be).

timholy:
Julia bug. It turns out that the combination of precompilation and limiting the compiler options seems to break coverage analysis. AFAICT its essentially limited to the calls that were made at the time the package was built, which means the "workload" executed during precompilation.

timholy:
Ah, it's https://github.com/JuliaLang/julia/issues/37059

PR #212, opened 2021-08-08
Caveat about false type instabilities, joining forces with `Infiltrator.jl`

goretkin:
I tried to emphasize the point that Cthulhu is a "static" tool. But (amazingly?!) you can use it from an infiltrator REPL, and get a more accurate understanding of type stability.

@baggepinnen @simeonschaub

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ By default,
 
 You can toggle between these with `o` and `w`.
 
-## Combining `Cthulhu` with `Infiltrator.jl`
+## Combining `Cthulhu` with execution
 
 Cthulhu has access only to "static" type information, the same information available to the Julia compiler and type inference.
 In some situations, this will lead to incomplete or misleading information about type instabilities.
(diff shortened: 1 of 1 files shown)


aviatesk:
Well, this is a general problem about how we can combine compile-time and runtime information.

The idea to integrate Cthulhu with Infilitrator is definitely valuable to mention, but technically we can also use Debugger.jl for example to start descending at the middle point of program execution.

So my idea is, to have a general wording "How to combine static and runtime information", and then mention the use of Infiltrator as a way to easily leverage runtime information ?

goretkin:
@aviatesk good point. I generalized the wording.

README.md
```diff
@@ -120,3 +120,100 @@ By default,
...
+Variables
+  #self#::Core.Const(bar)
+  x::Union{Float64, Int64}
+  y::Union{Float64, Int64}
+  z::Union{Float64, Int64}
+[...]
+```
+
+The text of `Union{Float64, Int64}`  will be in red, but it is likely that `bar` will be called via dynamic dispatch.
+`bar` will be a ["function barrier"](https://docs.julialang.org/en/v1/manual/performance-tips/#kernel-functions), and the types will be fully inferred within `bar`.
```
aviatesk, line 174:
```suggestion
The text of `Union{Float64, Int64}` will be colored in red indicating there are type-instabilities,
but they are unlikely to be problem in actual execution,
because `bar` here serves as a ["function barrier"](https://docs.julialang.org/en/v1/manual/performance-tips/#kernel-functions) and
`bar` will be called with fully concrete runtime types via dynamic dispatch.
```

aviatesk (approved):

aviatesk:
@goretkin sorry for being late to get back to you. Thanks so much for your contribution !

PR #214, opened 2021-08-08
combine with concrete interpretation

simeonschaub:
I was curious whether it would be possible to use JuliaInterpreter in order to not only get an abstract interpretation of a program, but also a concrete interpretation with the results displayed alongside the type inference results. This is currently quite naiv, only works on unoptimized IR and always interprets everything before code is displayed, which also means that the same code gets executed multiple times if one steps into calls, which is certainly not ideal if there is any mutable state. It does use these concrete types when descending into functions with multiple possible MethodInstances, which may also not always be the correct answer if a slot or ssavalue can have different types inside the same function call.

I am really just putting this up as a proof of concept. Maybe @timholy or @aviatesk have some ideas how we should present something like this to the user in a useful way. It would probably useful to have some kind of stepping, but I am worried that we will end up having to write a full debugger inside Cthulhu.

This currently depends on JuliaDebug/JuliaInterpreter.jl#491 and JuliaLang/julia#41828.
It can be invoked as `descend(sin, (Float64,); frameargs=Any[sin, 1.], optimize=false)` for example.


timholy:
Very intriguing! Let me start with https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/491 and then we can come back here when that's resolved.

simeonschaub:
I recently found this useful, so I updated it and I think it should be pretty much ready now. I only need to figure out how to properly test this.

PR #216, opened 2021-08-10
use consistent with InteractiveUtils syntax highlighting for llvm and native code

KristofferC:
I think the syntax highlighting that InteractiveUtils have is significantly nicer than the one provided by pygmentize. It is also nice to be consistent. And it avoids having to have the `pygmentize` library installed.
There is no julia syntax highlighter available from Base, if we want to use a pure Julia solution, there is https://github.com/JuliaDocs/Highlights.jl

Before:

Screenshot from 2021-08-10 15-14-54

Screenshot from 2021-08-10 15-15-16

After:

Screenshot from 2021-08-10 15-11-29

Screenshot from 2021-08-10 15-16-43

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -26,10 +26,9 @@ end
 
 # Options
 - `enable_highlighter::Bool`: Use command line `highlighter` to syntax highlight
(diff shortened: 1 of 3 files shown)


vchuravy:
The seqfault is now, but irreproducible outside CI so far...

PR #218, opened 2021-08-13
some type annotations and inferrability improvement

aviatesk:


src/interpreter.jl
```diff
@@ -91,8 +90,9 @@ function Compiler.inlining_policy(interp::CthulhuInterpreter)
 end
 
 function Compiler.finish!(interp::CthulhuInterpreter, caller::InferenceResult)
-    if isa(caller.src, OptimizationState)
-        opt = caller.src
+    src = caller.src
+    if isa(src, OptimizationState)
```
vchuravy, line 94: Run into something like this the other day, any way we can teach the compiler this?

aviatesk: https://github.com/JuliaLang/julia/pull/41199 for immutables, and EscapeAnalysis.jl + effect analysis for mutables.

PR #224, opened 2021-09-04
update to https://github.com/JuliaLang/julia/pull/42082

aviatesk:


src/interpreter.jl
```diff
@@ -84,12 +84,19 @@ end
 
 # branch on https://github.com/JuliaLang/julia/pull/41328
 @static if isdefined(Compiler, :is_stmt_inline)
-function Compiler.inlining_policy(interp::CthulhuInterpreter, @nospecialize(src), stmt_flag::UInt8)
```
aviatesk, line 87: The CIs are failing now because `1.8.0-DEV.472` doesn't seem to arrive at the nightly channel yet.
Still I'd like to drop the support for dev versions between https://github.com/JuliaLang/julia/pull/41328 and  `1.8.0-DEV.472`.

aviatesk:
@simeonschaub do you have any idea on why this test case now produces `Expr(:code_coverage_effect)` statements ... ?
https://github.com/JuliaDebug/Cthulhu.jl/blob/7ded0c99e09597bdf42444bbc3996dbbb67723d2/test/runtests.jl#L104-L116

simeonschaub:
Not sure why that changed, but you could use `PerformanceTestTools. @include` to work around it, as is already done here: https://github.com/JuliaDebug/Cthulhu.jl/blob/30ac9f5c69d118ad2e85e3068757b8df46b7d7f2/test/codeview.jl#L41

test/runtests.jl
```diff
@@ -99,29 +99,32 @@ let callsites = find_callsites_by_ftt(call_by_apply, Tuple{Tuple{Int}}; optimize
...
+#         end
+#     end
+#
+#     let
+#         (_,CI, _, _, _, _) = process(M.h, Tuple{Vector{Float64}})
+#         @test count(!isnothing, CI.stmts.inst) == 2
+#         stmt = CI.stmts.inst[end]
+#         @test isa(stmt, Core.ReturnNode) && !isdefined(stmt, :val)
+#     end
+# end
```
aviatesk, line 127: It turns out that this testset hasn't been ran so far. Running it in a separate process with `--check-bounds=auto --code-coverage=none` options will solve issues, but I will leave it for another PR.

PR #225, opened 2021-09-06
add option to show inference remarks

aviatesk:
- only enabled for unoptimized source
- show remarks aside of statements, rather than callsites
- would be mostly useful for understanding constant-prop' heuristics, but we can also see remarks emitted for e.g. recursive callsites

Would look like this:
Screen Shot 2021-09-06 at 21 01 54

### DIFF
--- a/src/ui.jl
+++ b/src/ui.jl
@@ -12,15 +12,15 @@ mutable struct CthulhuMenu <: TerminalMenus.ConfiguredMenu{TerminalMenus.Config}
     config::TerminalMenus.Config
 end
 
-function show_as_line(el, optimize::Bool, iswarn::Bool)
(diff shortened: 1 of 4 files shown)


src/ui.jl
```diff
@@ -118,7 +122,7 @@ function TerminalMenus.keypress(m::CthulhuMenu, key::UInt32)
     elseif key == UInt32('b')
         m.toggle = :bookmark
         return true
-    elseif key == UInt32('r') || key == UInt32('R')
```
aviatesk, line 121: One consideration of this PR would be that if we can remove <kbd>r</kbd> key from this Revise integration. @timholy can I hear your idea on this ?

timholy: Yeah, let's not have Revise steal 2 keystrokes. I'm fine with giving `r` to `:remarks`

PR #226, opened 2021-09-13
Fix typo in callsite showing

Keno:
Just a drive-by fix that caused Cthulhu to error for me.
Would be good to add whatever triggers this as a test case,
but I'm in the middle of some deep debugging, so I didn't get
the chance to look into this.

cc @simeonschaub @aviatesk

### DIFF
--- a/src/callsite.jl
+++ b/src/callsite.jl
@@ -240,7 +240,7 @@ function show_callinfo(limiter, ci::Union{MultiCallInfo, FailedCallInfo, Generat
     f = Compiler.argtype_to_function(ft)
     if f !== nothing
         name = "→ $f"
-    elseif f isa Union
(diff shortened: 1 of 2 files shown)


src/callsite.jl
```diff
@@ -243,7 +243,7 @@ function show_callinfo(limiter, ci::Union{MultiCallInfo, FailedCallInfo, Generat
     elseif f isa Union
```
simeonschaub, line 243:
```suggestion
    elseif tf isa Union
```

simeonschaub (approved):
Oops, that was my fault. It's definitely a bug since `f` is always `nothing` here.

Keno:
Add a testcase for both :)?

PR #233, opened 2021-10-05
fix printing of `PureCallInfo`

simeonschaub:
This errored before if the called object did not define `nameof`.


aviatesk (approved):
LGTM.

PR #234, opened 2021-10-05
add option to hide type annotations

simeonschaub:
The motivating case here is that Diffractor often generates complex and
deeply nested lattice elements, so hiding them can make the code easier
to read.


timholy:
We (meaning "Julia" writ large) have *got* to solve the type-printing problem. Is it really that hard? Why can't we "just" solve for the `env` using intersection against the function signature and print the decomposition of declared types and TypeVars?

simeonschaub:
Do you mean automatically dropping `TypeVar` for printing?

simeonschaub:
I'm not sure if it makes sense to implement this only for Cthulhu since we probably want to have this in Base as well, e.g. for stacktraces.

That's probably well out of scope for this PR though, so I would go with the solution here in the meantime.

vchuravy:
> I'm not sure if it makes sense to implement this only for Cthulhu 

But Cthulhu is great place to prototype this.

simeonschaub:
Sure, if someone wants to work on it! Ok to merge this first though?

vchuravy:
Yes, but we should do a UX sprint. I found that the limited printing of very large types has made Cthulhu less useful for me.

timholy:
> I'm not sure if it makes sense to implement this only for Cthulhu since we probably want to have this in Base as well, e.g. for stacktraces.

Agreed 100%. I'd say all the machinery should be in Base and just the UI part would be implemented by the environment.

And right, I didn't mean to imply that was a requirement for merging this.

PR #236, opened 2021-10-06
A more colourful menu

thazhemadam:
This PR adds colours to the menu options and also makes the colours for options that can be toggled (except for `debuginfo`) more expressive.

Before:

before

After:

after

### DIFF
--- a/src/ui.jl
+++ b/src/ui.jl
@@ -102,11 +102,17 @@ function usage(@nospecialize(view_cmd), annotate_source, optimize, iswarn, hide_
         active_option ? printstyled(io, c; bold=true, color=:green) : printstyled(io, c; color=:red)
     end
 
+    colorize(s::AbstractString; color::Symbol = :cyan) = stringify() do io
+        printstyled(io, s; color)
+    end
+
     io = IOBuffer()
     ioctx = IOContext(io, :color=>true)
 
-    println(ioctx, "Select a call to descend into or ↩ to ascend. [q]uit. [b]ookmark.")
-    print(ioctx, "Toggles: [",
+    println(ioctx,
+        colorize("Select a call to descend into or ↩ to ascend. [q]uit. [b]ookmark."; color=:blue))
+    print(ioctx,
+        colorize("Toggles"), ": [",
         colorize(iswarn, 'w'), "]arn, [",
         colorize(hide_type_stable, 'h'), "]ide type-stable statements, [",
         colorize(type_annotations, 't'), "]ype annotations, [",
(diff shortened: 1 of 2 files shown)


timholy (commented):
Personally I do like the new look! One concern is red/green colorblind users; it's the most common form, apparently about 1/12 of all men suffer from colorblindness https://aapos.org/glossary/color-blindness. Any chance we can choose a more colorblind-friendly set of colors?

thazhemadam:
I went with the red-green combo because it felt intuitive - green means yes and red means no; but I didn't take color blindness into account!

> Any chance we can choose a more colorblind-friendly set of colors?

Absolutely! Do you have any in mind or a personal preference from [these](https://davidmathlogic.com/colorblind) (or any other) palletes?

timholy:
I agree completely that for the non-colorblind red/green is indeed very intuitive, and that no other color combo quite matches it. But so many people are affected by some form of colorblindness that to me I think we have to account for their needs.

I am definitely not an expert and would defer to your or others' choices, but to me both the green/purple (lower left in "accessible palettes) or the blue/magenta (lower right) seem reasonably intuitive.

anandijain:
i'd love this feature! to be friendly to colorblind, maybe we could default the coloring to be as-is (off) and leave the config to another PR?

@thazhemadam thanks for doing this

fingolfin:
It's a pity this went stale :-(.

I agree we should take the needs of colorblind users in mind, but: my understanding is that someone who is e.g. red/green blind may not be able to differentiate between certain red and green colors (if they have the same brightness etc.), but they can still see them.

So wouldn't that mean for such a user, the effect would be that they don't see different colors -- so in effect exactly what we have no. I.e there would be no regression for them, just no improvement.

So while of course we should ultimately allow them to benefit, too (e.g. by making the colors configurable -- perhaps via [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl)), wouldn't it still make sense to merge this improvement now, and add the color preferences later?

serenity4:
> my understanding is that someone who is e.g. red/green blind may not be able to differentiate between certain red and green colors (if they have the same brightness etc.), but they can still see them.

The problem is, if we choose two colors with the same brightness, they won't even be able to distinguish between on/off. Fortunately, running the image from the issue description through https://www.color-blindness.com/coblis-color-blindness-simulator/, it seems that all of the different views agree on having the "green" brighter than the red, offering no confusion. In fact, it is much better than the current colors, where white and blue have a very similar brightness (e.g. see the image with original colors with the first two dichromatic views on the linked website). Of course feel free to correct me if this is not accurate, I am not at all an expert on the matter.

Not only does this PR with green/red colors make it more readable for most users, it also makes it actually readable for users with a certain color-blind type. I suggest we keep it that way.

thazhemadam:
I've rebased this, wioth some help from Claude. This should be good for review again now.

serenity4 (approved):
That's a small but great addition, thanks @thazhemadam!
It would also be nice to have custom colors for the various callsite annotations (like `< constprop >` etc), but I think this would better a follow-up PR if we want that.

For configuration, it is already possible to change the mapping between ANSI codes and the printed color in terminal settings, which I think is enough for now. We can always revisit the configuration later if needed.

timholy:
`debuginfo` has 3 settings, not two. Does this override the following? https://github.com/JuliaDebug/Cthulhu.jl/blob/935cca10e0e6207877849ebee82d4086814e954a/src/ui.jl#L96

PR #242, opened 2021-11-08
Add a native interpreter kwarg, forward method_table queries

maleadt:
Although this doesn't solve the fundamental problem of composing abstract interpreters (since we're still bypassing most of the native interpreter's logic here), by forwarding the `method_table` query to the now user-customizable native interpreter the GPU compiler can use this to reflect on GPU code using Cthulhu. Until we add other logic to the GPUInterpreter (e.g. custom inlining thresholds as in https://github.com/JuliaGPU/GPUCompiler.jl/pull/186), that is.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -471,22 +471,22 @@ function get_specialization(@nospecialize(F), @nospecialize(TT))
     return get_specialization(Base.signature_type(F, TT))
 end
 
-function mkinterp(native_interpreter::AbstractInterpreter, @nospecialize(args...))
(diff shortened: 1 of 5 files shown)


vchuravy (approved):

aviatesk (approved):
LGTM.

Though I'd like to name the keyword arguments (and other corresponding arguments/variables) as `interp`, since it would be more consistent with `code_typed`.

aviatesk:
Hm, yeah, maybe `interp′` is better to be `cthulhu` or something?
But LTGM anyway. Let me merge this as is.

PR #247, opened 2021-11-10
Add descend method for MethodInstance

chriselrod:
This is useful for descending into methodinstances you get from `SnoopCompile.@snoopi_deep`.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -165,6 +165,12 @@ Shortcut for [`descend_code_typed`](@ref).
 """
 const descend = descend_code_typed
 
+function descend(mi::MethodInstance; kwargs...)
+    interp = Cthulhu.CthulhuInterpreter()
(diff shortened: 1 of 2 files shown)


simeonschaub (approved):
Maybe add the same for `descend_code_warntype`? Otherwise LGTM.

chriselrod:
> Maybe add the same for `descend_code_warntype`? Otherwise LGTM.

Done.

vchuravy:
Can you add a test (probably around here https://github.com/JuliaDebug/Cthulhu.jl/blob/24562ef990ea840feb625c7ecdf6f3625a5fe935/test/runtests.jl#L498)

timholy (approved):
I added some tests; there's no point in dragging you through learning how to handle terminal tests, since they are a pain and not very relevant to the core functionality.

Thanks for doing this! I started adding this a half-dozen times and always got derailed.

PR #249, opened 2021-11-12
improve limited signature printing

simeonschaub:
We previously abbreviated all types using `T{…}` even if `T` didn't have
any type parameters. This fixes that and also simplifies the code
slightly.

### DIFF
--- a/src/callsite.jl
+++ b/src/callsite.jl
@@ -124,7 +124,7 @@ end
 get_mi(c::Callsite) = get_mi(c.info)
 
 # Callsite printing
-mutable struct TextWidthLimiter
+mutable struct TextWidthLimiter <: IO
     io::IO
     width::Int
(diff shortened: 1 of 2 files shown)


aviatesk (approved):

src/callsite.jl
```diff
@@ -124,7 +124,7 @@ end
 get_mi(c::Callsite) = get_mi(c.info)
 
 # Callsite printing
-mutable struct TextWidthLimiter
+mutable struct TextWidthLimiter <: IO
```
timholy, line 127: This is a pretty interesting change. In the long run I think `TextWidthLimiter` should be factored out into its own package (it's broadly useful). But I also recall, without remembering the details, that there were some subtleties about really supporting this as an `IO` subtype. Wish I could remember more.

PR #250, opened 2021-11-13
update `stmt_infos` after DCE

simeonschaub:
Previously this led to the statement infos not matching the IR anymore.
I am a bit unsure if we run DCE on unoptimized IR as well, in which case
we would have to update `infos` in the DCE pass as well.


aviatesk:
> I am a bit unsure if we run DCE on unoptimized IR as well, in which case
> we would have to update `infos` in the DCE pass as well.

DCE happens after abstract interpretation here: https://github.com/JuliaLang/julia/blob/89c137c1efa664a5999796f1a8b5d23d234cad5d/base/compiler/typeinfer.jl#L666-L675

But we collect unoptimized source by hooking into `finish(::InferenceState, ::CthulhuInterpreter)`, which is after that DCE happens, so the assertion you added should be good.

aviatesk (approved):

PR #261, opened 2022-01-12
Properly handle const-prop in opt/nonopt switching

Keno:
cc @ianatol if you could add a test for this.


Keno:
Almost forgot about this. @ianatol can you push it through?

timholy:
With SnoopCompile 2.9 out, I want to see how the integration test is doing (close/reopen cycle).

aviatesk:
Will be fixed by #269.

Issue #259, opened 2022-01-11
Remarks printing is very slow


Keno:
The remarks printing is currently quadratic, because of the following code:
https://github.com/JuliaDebug/Cthulhu.jl/blob/5e88b2419eb06d36e11d30f4487890b5f28a3d4d/src/codeview.jl#L169-L182
This should be sorted in a pre-processing pass and be iterated one at a time or at the very least use a sorted lookup.

cc @ianatol

Keno:
For reference the function I'm looking at has 15k statements and 12k remarks, so iterating through all of them is quite expensive.

PR #267, opened 2022-02-14
sort remarks for n*log(n) printing

oscardssmith:
fixes https://github.com/JuliaDebug/Cthulhu.jl/issues/259

### DIFF
--- a/src/codeview.jl
+++ b/src/codeview.jl
@@ -167,14 +167,14 @@ function cthulhu_typed(io::IO, debuginfo::Symbol,
         Returns(nothing)
     end
     if !isnothing(remarks)
+        sort!(remarks)
         function postprinter(io::IO, @nospecialize(typ), used::Bool)
             _postprinter(io, typ, used)
             haskey(io, :idx) || return
             idx = io[:idx]::Int
-            for (pc, remark) in remarks
-                if pc == idx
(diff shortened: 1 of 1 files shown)


oscardssmith:
I don't think the failures on nightly are my fault, but I'm not 100% sure.

aviatesk:
> One question: Is there 1 remark per pc? If not, this is slightly broken and it needs to iterate a bit.

yes there can be multiple remarks per statement

src/codeview.jl
```diff
@@ -167,14 +167,15 @@ function cthulhu_typed(io::IO, debuginfo::Symbol,
...
         function postprinter(io::IO, @nospecialize(typ), used::Bool)
             _postprinter(io, typ, used)
             haskey(io, :idx) || return
             idx = io[:idx]::Int
-            for (pc, remark) in remarks
-                if pc == idx
-                    printstyled(io, ' ', remark; color=:light_black)
-                end
+            firstind = searchsortedfirst(remarks, idx=>"")
+            for i in firstind:lastindex(remarks)
```
aviatesk, line 176: can we further limit this iteration space to `firstind:searchsortedfirst(remarks, idx+1=>"")`?

aviatesk: ah I see.

PR #269, opened 2022-02-15
show optimized constant-prop'ed source when switched from unoptimized constant-prop'ed source

aviatesk:


ianatol:
Ok, after some messing around with this, I'm happy to call this good. The situation where we encountered this bug was when we had modified `mkinterp` and `do_typeinf!` as follows, to make an entry point for doing const prop:

```julia
function do_typeinf!(interp::Cthulhu.CthulhuInterpreter, mi::Core.Compiler.MethodInstance, argtypes = nothing)
    if argtypes !== nothing
        arginfo = Core.Compiler.ArgInfo(nothing, argtypes)
        result = Core.Compiler.InferenceResult(mi, (arginfo, nothing), false)
    else
        result = Core.Compiler.InferenceResult(mi)
    end
    frame = Core.Compiler.InferenceState(result, #=cache=# :global, interp)::Core.Compiler.InferenceState
    Core.Compiler.typeinf(interp, frame)
    return result
end

function mkinterp_with_const_args(interp::Core.Compiler.AbstractInterpreter, argtypes=nothing, @nospecialize(args...))
    interp′ = Cthulhu.CthulhuInterpreter(interp)
    mi = Cthulhu.get_specialization(args...)
    do_typeinf!(interp′, mi, argtypes)
    (interp′, mi)
end
```
Doing this form of const prop, there were cases in our example where `key` wasn't in `interp.unopt`, but `override` was. However, with this change and especially the effects/consteval changes in master, I don't think its worth the time to come up with a very convoluted test case when we know this bug doesn't occur with normal inference anyway (where we always infer the method instance first). I don't see a reason not to check `interp.unopt` for `override` like in #261 though

ianatol:
Thoughts? @aviatesk

aviatesk:
Whether or not if we pass `override::InferenceResult` (a constprop' result), there always should be an associated `MethodInstance` with it (i.e. override.linfo`) and so `opt`/`unopt` are supposed to have that.
I guess your bug was caused probably because you used different `CthulhuInterpreter`s betwen the additional constprop' trial and the main `descend` routine, and then it is supposed not to work really.

Well, it would be interesting if we have an entry point with extended lattice elements.
Maybe we want to work on it first and figure out #261 is really needed to support such an entry.

PR #272, opened 2022-02-28
propagate TTY displaysize

aviatesk:
Currently `displaysize` configuration of TTY terminals are not respected, and so the inline source location printing appears to be a bit weird:
> before
Screen Shot 2022-02-28 at 12 21 11

With this PR, we get a reasonable a location printing:
> after
Screen Shot 2022-02-28 at 12 20 44

### DIFF
--- a/src/codeview.jl
+++ b/src/codeview.jl
@@ -113,8 +113,7 @@ function cthulhu_typed(io::IO, debuginfo::Symbol,
     debuginfo = IRShow.debuginfo(debuginfo)
     lineprinter = __debuginfo[debuginfo]
     rettype = ignorelimited(rt)
-    iolim = IOContext(io, :limit=>true)
-    lambda_io = iolim
+    lambda_io = IOContext(io, :limit=>true)
 
     if isa(src, Core.CodeInfo)
         # we're working on pre-optimization state, need to ignore `LimitedAccuracy`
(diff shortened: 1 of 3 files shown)


aviatesk:
Going to merge this and tag a new patch version once CI runs correctly.

src/codeview.jl
```diff
@@ -148,7 +147,7 @@ function cthulhu_typed(io::IO, debuginfo::Symbol,
         maxcost = Core.Compiler.statement_costs!(cst, code, src, Any[mi.sparam_vals...], false, params)
         nd = ndigits(maxcost)
         _lineprinter = lineprinter(src)
-        function preprinter(io, linestart, idx)
+        preprinter = function (io, linestart, idx)
```
simeonschaub, line 150: Was there a reason for this change? I generally prefer the former since it will give more informative backtraces

src/ui.jl
```diff
@@ -47,39 +47,42 @@ end
...
-        colorize(iotmp, view_cmd === cthulhu_source, 'S'), "]ource code, [",
-        colorize(iotmp, view_cmd === cthulhu_ast, 'A'), "]ST, [",
-        colorize(iotmp, view_cmd === cthulhu_typed, 'T'), "]yped code, [",
-        colorize(iotmp, view_cmd === cthulhu_llvm, 'L'), "]LVM IR, [",
-        colorize(iotmp, view_cmd === cthulhu_native, 'N'), "]ative code")
+        colorize(view_cmd === cthulhu_source, 'S'), "]ource code, [",
+        colorize(view_cmd === cthulhu_ast, 'A'), "]ST, [",
+        colorize(view_cmd === cthulhu_typed, 'T'), "]yped code, [",
+        colorize(view_cmd === cthulhu_llvm, 'L'), "]LVM IR, [",
+        colorize(view_cmd === cthulhu_native, 'N'), "]ative code")
```
simeonschaub, line 85: Won't this just always print to stdout? I don't think we want that

aviatesk: Nah, now `stringify` function has been refactored so that it always creates a new `IOBuffer` object to generate a string from a given printing function

src/codeview.jl
simeonschaub: That should only effect the case where we actually define different methods, but I see your point

src/ui.jl
simeonschaub: Ah, ok! I was just confused since I thought `colorize` actually printed to the io instead of just outputting a string.

simeonschaub (approved):
LGTM

src/codeview.jl
aviatesk: ah, I see. Then let's try to keep more informative backtraces.

PR #273, opened 2022-03-03
update to JuliaLang/julia#44389

aviatesk:
The `method_table` interface changed by that PR.
Cthulhu doesn't need to overload it from the first place.

### DIFF
--- a/src/interpreter.jl
+++ b/src/interpreter.jl
@@ -37,15 +37,14 @@ CthulhuInterpreter(interp::AbstractInterpreter=NativeInterpreter()) =
     )
 
 import Core.Compiler: InferenceParams, OptimizationParams, get_world_counter,
-    get_inference_cache, code_cache, method_table,
-    WorldView, lock_mi_inference, unlock_mi_inference, InferenceState
+    get_inference_cache, code_cache, lock_mi_inference, unlock_mi_inference
+import Core.Compiler: InferenceState, WorldView
 using Base: @invoke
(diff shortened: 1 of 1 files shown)


src/interpreter.jl
```diff
@@ -37,15 +37,14 @@ CthulhuInterpreter(interp::AbstractInterpreter=NativeInterpreter()) =
...
-    WorldView, lock_mi_inference, unlock_mi_inference, InferenceState
+    get_inference_cache, code_cache, lock_mi_inference, unlock_mi_inference
+import Core.Compiler: InferenceState, WorldView
 using Base: @invoke
 
 Compiler.InferenceParams(interp::CthulhuInterpreter) = InferenceParams(interp.native)
 Compiler.OptimizationParams(interp::CthulhuInterpreter) = OptimizationParams(interp.native)
 Compiler.get_world_counter(interp::CthulhuInterpreter) = get_world_counter(interp.native)
 Compiler.get_inference_cache(interp::CthulhuInterpreter) = get_inference_cache(interp.native)
-Compiler.method_table(interp::CthulhuInterpreter, sv::InferenceState) = method_table(interp.native, sv)
```
vchuravy, line 48: We don't need to forward it to the inner AbsInt? The goal was to be able to wrap a GPUCompiler AbsInt eventually

aviatesk: Thanks, see #274.

PR #281, opened 2022-03-22
Store effects in OptimizationSource

ianatol:
@Keno pointed out to me that https://github.com/JuliaDebug/Cthulhu.jl/blob/master/src/Cthulhu.jl#L343 was broken. To fix, we want to store effects in `OptimizationSource` and access them from there instead of the current, faulty way.

### DIFF
--- a/src/interpreter.jl
+++ b/src/interpreter.jl
@@ -15,6 +15,15 @@ struct OptimizedSource
     ir::IRCode
     src::CodeInfo
     isinlineable::Bool
+    effects::Effects
+end
+
+maybe_create_optsource(@nospecialize(a), ::Effects) = a
+
+function maybe_create_optsource(st::OptimizationState, effects::Effects)
+    ir = st.ir
+    ir === nothing && return st
+    return OptimizedSource(ir, st.src, st.src.inlineable, effects)
 end
 
 const Remarks = Vector{Pair{Int, String}}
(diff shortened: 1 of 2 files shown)


src/interpreter.jl
```diff
@@ -76,25 +85,19 @@ function Compiler.add_remark!(interp::CthulhuInterpreter, sv::InferenceState, ms
...
         valid_worlds::WorldRange, @nospecialize(inferred_result))
-    if isa(inferred_result, OptimizationState)
-        opt = inferred_result
-        ir = opt.ir
-        if ir !== nothing
-            return OptimizedSource(ir, opt.src, opt.src.inlineable)
-        end
-    end
-    return inferred_result
+    return maybe_create_optsource(inferred_result, Effects())
```
Keno, line 100: In https://github.com/JuliaLang/julia/pull/44660, I modified this function to pass in the effects. Maybe pull that part out of that PR and submit it separately, so we can use it here?

ianatol:
Should be good to go now that https://github.com/JuliaLang/julia/pull/44777 is merged

simeonschaub:
Does that PR need to be backported? Otherwise, Cthulhu would be broken on 1.8, right?

ianatol:
Hmm yeah, we would want to backport it for that effects printing to properly work on 1.8. It would also be a bit cleaner to just check `EFFECTS_ENABLED` rather than the version number.

ianatol:
Rebased with #285

src/interpreter.jl
```diff
@@ -15,6 +15,15 @@ struct OptimizedSource
     ir::IRCode
     src::CodeInfo
     isinlineable::Bool
+    effects::Effects
+end
+
+maybe_create_optsource(@nospecialize(a), _) = a
```
aviatesk, line 21:
```suggestion
maybe_create_optsource(@nospecialize(a), ::Effects) = a
```

aviatesk (approved):

PR #282, opened 2022-03-25
fix `OCCallInfo` construction, add test for OC callsite

aviatesk:


aviatesk:
@ianatol can you take a review on this change?

ianatol:
Was out sick last week so sorry for the delay, but LGTM!

PR #286, opened 2022-03-30
Improve get_effects functionality, add tests

ianatol:
This adds some functions and tests to at least ensure that `get_effects` doesn't error for each type of callsite

Accuracy for each likely still needs work

CC @aviatesk

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -46,8 +46,14 @@ isordered(::Type{T}) where {T<:AbstractDict} = false
     callsites = find_callsites_by_ftt(iterate, (Base.IdSet{Any}, Union{}); optimize=false)
     @test callsites[1].info isa Cthulhu.ConstPropCallInfo
 
-    # Broken stuff in Julia
-    @test_broken find_callsites_by_ftt(Core.Compiler._limit_type_size, Tuple{Any, Type{Any}, Core.SimpleVector, Int, Int})  # ssair/ir.jl bug
+    if Cthulhu.EFFECTS_ENABLED
+        effects = Cthulhu.get_effects(callsites[1].info)
+        @test !Core.Compiler.is_concrete_eval_eligible(effects)
+        @test !Core.Compiler.is_consistent(effects)
+        @test Core.Compiler.is_effect_free(effects)
+        @test Core.Compiler.is_nothrow(effects)
+        @test Core.Compiler.is_terminates(effects)
+    end
 end
 
 @testset "Expr heads" begin
@@ -138,6 +144,7 @@ let callsites = find_callsites_by_ftt(f_matches, Tuple{Any, Any}; optimize=false
     @test length(callsites) == 1
     callinfo = callsites[1].info
     @test callinfo isa Cthulhu.MultiCallInfo
+    Cthulhu.EFFECTS_ENABLED && @test Cthulhu.get_effects(callinfo) |> Core.Compiler.is_concrete_eval_eligible
     io = IOBuffer()
     Cthulhu.show_callinfo(io, callinfo)
     @test occursin(r"→ g_matches\(::Any, ?::Any\)::Union{Float64, ?Int\d+}", String(take!(io)))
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -46,6 +46,8 @@ isordered(::Type{T}) where {T<:AbstractDict} = false
     callsites = find_callsites_by_ftt(iterate, (Base.IdSet{Any}, Union{}); optimize=false)
     @test callsites[1].info isa Cthulhu.ConstPropCallInfo
 
+    @test Cthulhu.get_effects(callsites[1].info) isa Cthulhu.Effects
```
aviatesk, line 49: Maybe we can do this check within `find_callsites_by_ftt` instead, which would then function as a basic test for `CallInfo` interface.

At each test case, we may want to focus on the correctness of collected effects information, rather than checking the basic interface requirement.

ianatol: Done (as far as we can for now)

test/runtests.jl
```diff
@@ -345,6 +382,10 @@ end
         callsite = only(callsites)
         info = callsite.info
         @test isa(info, Cthulhu.InvokeCallInfo)
+        if Cthulhu.EFFECTS_ENABLED
+            effects = Cthulhu.get_effects(info)
+            @test_broken Core.Compiler.is_total(effects) # don't have effects from InvokeCallInfo yet
```
ianatol, line 387: I think this may require #281 but not sure. The comment is wrong though, we do grab effects from `InvokeCallInfo`

test/runtests.jl
```diff
@@ -46,6 +46,14 @@ isordered(::Type{T}) where {T<:AbstractDict} = false
     callsites = find_callsites_by_ftt(iterate, (Base.IdSet{Any}, Union{}); optimize=false)
     @test callsites[1].info isa Cthulhu.ConstPropCallInfo
 
+    if Cthulhu.EFFECTS_ENABLED
+        effects = Cthulhu.get_effects(callsites[1].info)
+        @test !Core.Compiler.is_consistent(effects)
+        @test Core.Compiler.is_effect_free(effects)
+        @test Core.Compiler.is_nothrow(effects)
+        @test Core.Compiler.is_terminates(effects)
```
ianatol, line 54: More relevant for Base, but we should maybe have an interface for testing these "abnormal" effects more concisely, like the `Effects` constructor

aviatesk: Do you mean we need something other than `!is_concrete_elibible` (`!is_total_or_error`, previously)?

test/runtests.jl
```diff
@@ -345,6 +374,7 @@ end
         callsite = only(callsites)
         info = callsite.info
         @test isa(info, Cthulhu.InvokeCallInfo)
+        Cthulhu.EFFECTS_ENABLED && @test_broken Cthulhu.get_effects(info) |> Core.Compiler.is_total  # don't have effects from InvokeCallInfo yet
```
aviatesk, line 377: Can't we try to get effects from `Core.invoke` callsite? Now I think we are just passing `Effects()` explicitly in such case.
https://github.com/JuliaDebug/Cthulhu.jl/blob/8905460b785a67dbfcb759a37687429756f3b9ec/src/reflection.jl#L157

test/runtests.jl
```diff
@@ -158,6 +164,12 @@ end
         @test length(callsites) == 1
         ci = first(callsites).info
         @test isa(ci, Cthulhu.UncachedCallInfo)
+        if Cthulhu.EFFECTS_ENABLED
+            effects = Cthulhu.get_effects(ci)
+            @test !Core.Compiler.is_consistent(effects)
+            @test Core.Compiler.is_effect_free(effects)
+            @test !Core.Compiler.is_nothrow(effects)
```
aviatesk, line 171: Maybe `@test !Core.Compiler.is_terminates(effects)` also? Mutually recursive function calls usually prohibit a proof of termination.

aviatesk (approved):
Nicely done, thanks @ianatol !

PR #288, opened 2022-04-19
update to upstream rename refactoring

aviatesk:
xref: JuliaLang/julia#44888


src/reflection.jl
```diff
@@ -140,11 +140,21 @@ function process_info(interp, @nospecialize(info), argtypes::ArgTypes, @nospecia
         return mapany(enumerate(info.results)) do (i, result)
             if isnothing(result)
                 infos[i]
+            elseif (@static isdefined(Compiler, :ConcreteResult) && true) && isa(result, Compiler.ConcreteResult)
```
vchuravy, line 143: I prefer VERSION checks over `isdefined` checks, since those are more fragile to renames and reuse of names over times

PR #293, opened 2022-05-24
Update with upstream codeinfo changes

ianatol:
This updates Cthulhu with https://github.com/JuliaLang/julia/pull/45378


ianatol:
Not sure if this is really necessary because we ended up providing deprecated access to the same field in https://github.com/JuliaLang/julia/pull/45378, but I guess it uses the new accessors when available.

PR #296, opened 2022-06-05
Allow the default state of toggles in `@descend` to be configured and saved persistently

brenhinkeller:
This PR (1) allows the default state of various toggles in `@descend`/`_descend` to be set in `Cthulhu.CONFIG` and (2) allows that configuration to be persistently saved between sessions using Preferences.jl

Example:
```
using Cthulhu

# Customize some defaults
Cthulhu.CONFIG.enable_highlighter = true
Cthulhu.CONFIG.debuginfo = :none

# Save to LocalPreferences.toml using Preferences.jl
# Will be automatically read next time you `using Cthulhu`
Cthulhu.save_config!(Cthulhu.CONFIG)
```

### DIFF
--- a/src/preferences.jl
+++ b/src/preferences.jl
@@ -3,15 +3,14 @@
 ```julia
 save_config!(config::CthulhuConfig=CONFIG, force=true)
 ```
-If Preferences.jl is loaded, save a Cthulhu.jl configuration `config`
-(by default, `Cthulhu.CONFIG`) to your `LocalPreferences.toml` file.
+Save a Cthulhu.jl configuration `config` (by default, `Cthulhu.CONFIG`) to your
+`LocalPreferences.toml` file using Preferences.jl.
 
 The saved preferences will be automatically loaded next time you `using Cthulhu`
-if Preferences.jl is loaded.
 
 ## Examples
 ```julia
-julia> using Cthulhu, Preferences
+julia> using Cthulhu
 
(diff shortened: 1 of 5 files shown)


src/preferences.jl
```diff
@@ -0,0 +1,57 @@
...
+
+julia> Cthulhu.CONFIG.debuginfo = :none     # Customize some defaults
+:none
+
+julia> Cthulhu.save_config!(Cthulhu.CONFIG) # Will be automatically read next time you `using Cthulhu`
+```
+"""
+function save_config!(config::CthulhuConfig=CONFIG, force=true)
+    cthulhu_id = UUID("f68482b8-f384-11e8-15f7-abe071a5a75f")
+    set_preferences!(cthulhu_id,
```
vchuravy, line 26:
```suggestion
    @set_preferences!(
```

vchuravy (approved):
Besides maybe using the macro versions this looks great!

vchuravy:
Could you also add a simple test?

src/preferences.jl
brenhinkeller: Ah yeah, I couldn't figure out how to get the macros when doing the "use without adding as dependency" trick, but we're not doing that any more so might as well go for the macros!

brenhinkeller:
I think that should do it!

brenhinkeller:
Awesome, thanks!

PR #303, opened 2022-07-17
Refactor Cthulhu to allow custom callinfo for external packages

Keno:
This is intended to allow Diffactor to provide a Cthulhu interfaces
for navigating between different differentiation levels. The core
interface is the introduction of `Cthulhu.AbstractCursor`, which
downstream packages can extend to provide additional information
beyond an mi to identify the frame that's currently being displayed.
These interfaces will probably need to evolve a bit, but I think this
is a reasonable start.

Also strengthen the tests a bit to make failures more obvious and
easier to debug.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -3,6 +3,7 @@
 
 Required overloads:
 - `Cthulhu.lookup(interp::AbstractInterpreter, curs::AbstractCursor, optimize::Bool)`
+- `Cthulhu.lookup_constproped(interp::AbstractInterpreter, curs::AbstractCursor, override::InferenceResult, optimize::Bool)`
 - `Cthulhu.get_mi(curs::AbstractCursor) -> MethodInstance`
 - `Cthulhu.update_cursor(curs::AbstractCursor, mi::MethodInstance)`
 - `Cthulhu.navigate(curs::AbstractCursor, callsite::Callsite) -> AbstractCursor`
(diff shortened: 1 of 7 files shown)


vchuravy (approved):
Maybe we eventually need to factor this out to a Interface package?

Then we could move the GPU integration to GPUCompiler

Keno:
> Maybe we eventually need to factor this out to a Interface package?

Yes, I think we'll want to go this way, but we should probably refine the interface first.

src/interface.jl
```diff
@@ -0,0 +1,36 @@
...
+    mi::MethodInstance
+end
+
+function update_mi(curs::CthulhuCursor, mi::MethodInstance)
+    CthulhuCursor(mi)
+end
+
+# This method is optional, but should be implemented if there is
+# a sensible default cursor for a MethodInstance
+AbstractCursor(interp::CthulhuInterpreter, mi::MethodInstance) =
```
aviatesk, line 12:
```suggestion
AbstractCursor(interp::AbstractInterpreter, mi::MethodInstance) =
```

aviatesk: Should make it optional.

src/interface.jl
```diff
@@ -0,0 +1,36 @@
+abstract type AbstractCursor; end
+struct CthulhuCursor <: AbstractCursor
+    mi::MethodInstance
+end
+
+function update_mi(curs::CthulhuCursor, mi::MethodInstance)
+    CthulhuCursor(mi)
+end
```
aviatesk, line 8:
```suggestion
update_mi(curs::AbstractCursor, ::MethodInstance) = error("""
missing `$AbstractCursor` API:
`$(typeof(curs))` is required to implement the `$update_mi(::$(typeof(curs)), ::MethodInstance) -> $(typeof(curs))` interface.
""")
update_mi(curs::CthulhuCursor, mi::MethodInstance) = CthulhuCursor(mi)
```

src/interface.jl
```diff
@@ -0,0 +1,36 @@
...
+# This method is optional, but should be implemented if there is
+# a sensible default cursor for a MethodInstance
+AbstractCursor(interp::CthulhuInterpreter, mi::MethodInstance) =
+    CthulhuCursor(mi)
+
+function get_optimized_code(interp::CthulhuInterpreter,
+                                curs::CthulhuCursor)
+    return interp.opt[curs.mi].inferred
+end
+
```
aviatesk, line 19:
```suggestion
```
Never used.

Keno:
I have some additional changes on top of this, but I think rather than ping-ponging this PR back and forth, I'm gonna merge this as is and then do my additional changes in a quick separate PR.

PR #305, opened 2022-07-22
Respect const prop info in `OCCallInfo`/`InvokeCallInfo`

Keno:
Noticed while trying to debug Diffractor and ending up in the non-constprop OC invocation. @aviatesk could you finish this up and add a test case? I think invoke probably needs the same treatment also.


aviatesk:
Okay, this should be ready. Going to merge on successful CI.

PR #318, opened 2022-08-30
fix v1.7 compat

aviatesk:


timholy:
Thanks for the quick fix! (Didn't bother me, but I happened to stumble across it so...)

PR #323, opened 2022-09-06
adjust to JuliaLang/julia#46574

aviatesk:


test/runtests.jl
```diff
@@ -9,8 +9,13 @@ using Test, PerformanceTestTools
...
 
-    @testset "test_irshow.jl" begin
-        PerformanceTestTools.@include("test_irshow.jl")
+    # TODO enable this test on nightly
+    @static if VERSION < v"1.9.0-DEV.1295"
+        @testset "test_irshow.jl" begin
+            PerformanceTestTools.@include("test_irshow.jl")
+        end
+    else
+        @info "skipped test_irshow.jl"
```
aviatesk, line 18: https://github.com/JuliaLang/julia/pull/46629 changed the inlining of const `GlobalRef`, that caused the failures of the `irshow` test cases.
We should rather enable this test only on nightly, and maybe also want to enable tests for effects-related rendering.

PR #334, opened 2022-12-01
Use literal instead of undefined `optimize` var

### LINKED ISSUE #332: UndefVarError: optimize not defined
https://github.com/JuliaDebug/Cthulhu.jl/blob/a870c8aa153f6cb69d1e5e17784b43e4f7f67e25/src/Cthulhu.jl#L351-L368

On line 366, `optimize` is undefined, so this branch is broken.

simsurace:
Fixes #332.


simsurace:
Friendly bump :)

jishnub:
Gentle bump @aviatesk

Keno:
Would be good to get a test for a case that triggers this, so we can be sure it doesn't regress.

PR #336, opened 2022-12-12
add a `[v]iew` config option to always call `edit` on descend/ascend

anandijain:
this makes interactive IDE usage a lot nicer IMO and aligns more closely to how i used to descend, which was basically continually option clicking function calls within VSCode and manually picking the right dispatch.

i couldn't get the behavior for adding it as a toggle to work quite right though, which is why i have the `view_always` default to true currently.


vchuravy:
Continued in #485

PR #337, opened 2023-01-10
Warn color for small unions

jishnub:
This PR uses the color scheme from `@code_warntype` in `@descend_code_warntype`. Currently, the `%` signs for non-inferred return types are always displayed in red, but these are largely distractions for small unions.

After this,

Screenshot from 2023-01-10 20-15-45

The color of the `%` in `%3` is yellow instead of red.

Tests pass locally on v1.8, but fail on v1.9.0-beta2 with the same errors as on master, so I don't think this PR introduces errors.


aviatesk (approved):
I like this! Thanks so much for your contribution!

PR #339, opened 2023-01-11
read runtime preferences at runtime

vtjnash:
Preferences are not supposed to be read during compilation, since it invalidates the compilation cache.


vchuravy:
> Preferences are not supposed to be read during compilation, since it invalidates the compilation cache.

I would say, these preferences don't need to be compiletime preferences.

vchuravy (approved):

PR #343, opened 2023-01-21
Close pipe during precompilation

### LINKED ISSUE #342: Precompilation hangs on nightly
```julia
julia> using Cthulhu
[ Info: Precompiling Cthulhu [f68482b8-f384-11e8-15f7-abe071a5a75f]

[pid 26619] waiting for IO to finish:
 TYPE[FD/PID]       @UV_HANDLE_T->DATA
 pipe[11]           @0x556b98794560->0x7f6cfda14e20
```
and this message keeps repeating at intervals until you press Ctrl-C.

Today has been busy and I haven't even looked at fixing this yet, other than knowing it's another case of https://github.com/JuliaLang/julia/pull/46571 (CC @vtjnash). In the case of [GtkObservables](https://github.com/JuliaGizmos/GtkObservables.jl/pull/46) it was a Timer that I needed to make sure got properly `close`d when the object was destroyed (i.e., a good change to GtkObservable's code, since there's nothing desirable about having a timer keep running after its corresponding window closes).

timholy:
Fixes #342

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -813,6 +813,7 @@ if ccall(:jl_generating_output, Cint, ()) == 1
     term = REPL.Terminals.TTYTerminal("dumb", input.out, IOBuffer(), IOBuffer())
     write(input.in, 'q')
     descend(gcd, (Int, Int); terminal=term)
+    close(input)
     nothing
 end
 


timholy:
Failures on nightly seem independent of this PR.

vtjnash:
The error here is possibly a more sinister one, that a lack of a reader means this code probably does not run to completion. Closing the Pipe merely destroys the error channel too for the REPL

PR #345, opened 2023-02-05
RFC/WIP: map type annotations to source text

timholy:
This is an early-stage effort to make Cthulhu "newbie-friendly" by printing type-annotations against the raw source code. Timing-wise, there are several factors that led me to tackle & submit this now: 

- I'm submitting very early because I suspect @aviatesk has given this some thought and I didn't want to spend too much time on it before providing opportunities for discussion
- [JuliaSyntax](https://github.com/JuliaLang/JuliaSyntax.jl) appears to be converging and partially solves https://github.com/JuliaLang/julia/issues/31162
- Julia 1.9 seems likely to increase enthusiasm for hunting invalidations, which is currently tricky if you don't read lowered/type-inferred code
- (most proximally) a course I'm teaching will hit type-inference in a couple of weeks and it would be lovely to have something for the class to play with.
- overall, I regard this as one of the (relatively few) significant usability issues we have to cross before Julia becomes ready for "everyone"

First a small (and very imperfect) demo is in order:

image

(After this there is "cruft" from the fact that this source view has not been fully integrated so that it supersedes the current infrastructure.)

To hint at the option to suppress type-stable types,

image

Here it's only being applied to the signature, since very little else is type-stable in this example.

One really big caveat: I'm not sure this can be done *properly* until JuliaSyntax also handles lowering. The ugliest part of this by far is matching parsed expressions to their type-inferred equivalents. I'd particularly appreciate hearing thoughts on this issue.

In the long term future, it might be nice to allow the cursor to traverse the source text and enter into particular calls from there, without needing to print the list of calls.

CC @c42f


timholy:
Still very crude, but perhaps already starting to look promising. From
```julia
@descend optimize=false iswarn=true annotate_source=true hide_type_stable=true Base.var"#mapslices#193"(2,mapslices, sum,A)
```
(which on my version of Julia is the "main" call from `mapslices(sum, A; dims=2)` where `A::Array{Int,3}`),

image

timholy:
This is looking far more successful than I expected. I believe it's now working well enough that this PR is now heading for completion:

- [ ] needs tests
- [ ] needs proper integration with Cthulhu. I'm planning to make this the **default** mode of Cthulhu (so get your complaints in soon 😉). People who use Cthulhu primarily to better understand the compiler can of course use Preferences to pick a different default mode.
- [ ] needs documentation

The one part that I think won't happen soon is in-source call navigation, as that involves fancy REPL features we currently lack. For now I think we'll have to live with the current SSA-based call list. We might at least be able to strip the SSAs and print line numbers which could be matched against line numbers printed alongside the source.

CC @pfitzseb in case there's any of this that should be done mindfully of VS Code.

vchuravy:
I think this is a fabolous idea and effort. There are a few other low hanging UX efforts that could go well with this (like using Term.jl or something similar to implement paging for source code).

The bigger question here for me is. Should this be the default mode? I am also feel that this transform Cthulhu to something else entirely and it might be better called Cuddulu or FluffyCthulhu. I love the name Cthulhu but it was also in part choosen to convey the notion of "all hope is lost, but I will persist in my quest for knowledge", that surrender of sanity in pursuit of elder wisdom xD

vchuravy:
The other question is, can we do a medium thing? I don't find it all to useful to just show the `Any` I would certainly want to know the types of the call that led to `Any`

timholy:
We could certainly split it out. We'd need a common core that implements all the wonderful work that's been done to ensure that inference is handled correctly. Would Cthulhu-proper then be much more than a UI package?

>  I don't find it all to useful to just show the Any I would certainly want to know the types of the call that led to Any

At present, the source will show `::Any` but the list of calls below (ones you can descend into) include the argtypes. Or you can dive in and see more detail. Long arg lists and complicated types of course are a bit troublesome (there may not be room on a single line to print all of them), in those cases descending would probably be the only option.

In the long run, though, I had imagined a "cursor" over the source code, and you can dive into calls simply by moving the cursor to the right call and hitting `Enter`. That would obscure information about the argtypes. Perhaps `<space>` could be configured to display the argtypes of the call?

timholy:
> I love the name Cthulhu but it was also in part choosen to convey the notion of "all hope is lost, but I will persist in my quest for knowledge", that surrender of sanity in pursuit of elder wisdom

At the risk of ruining the fun of a lovely joke, I wonder how many beginning coders might feel that way even about the "fluffy" mode this implements. "Why does Julia have types at all, it's so annoying?!?" That is, to sophisticated developers like you, the two/three/four/five-language problem that you need to be able to read and understand Julia+lowered Julia+type-inferred Julia+IRCode+LLVM+native code seems daunting, but for complete newbies even the notion of tracing types through a program may be a level of complexity that just about breaks them, even when we try to help them as much as we can. If there's more than a grain of truth in that, then perhaps the name Cthulhu is still appropriate, considering the audiences.

"Cthulhu: think it's easy? Wait until you try to gain the *next* level of insight!"

pfitzseb:
Since VS Code was brought up: https://github.com/julia-vscode/LanguageServer.jl/pull/1077 is somewhat relevant.

As always, with the current design of the language server, there's a bit of a tension between wanting to get type info statically from the source code, but having a full REPL context getting us much further. JET.jl does have integration for "runtime diagnostics"; we could conceivably do something similar for Cthulhu as well.

vchuravy:
Yeah I sometimes wonder if Cthulhu as a name is not just another barrier to entry. Hard to spell, hard to pronounce, comes from a fantasy canon with questionable pedigree... It captured very well my emotional state at the time of first draft, but it might not be the best name,

timholy:
I do see your point. Splitting the fluffy version out would be very easy and provide an opportunity for a rename, reserving the scarier mode for deep analysis.

I think it is fitting & best to have you make the call, so just let me know when you've decided.

c42f (commented):
This looks cool! So great to see `JuliaSyntax` in use :)

Some of the parts you're trying to use here are very under-documented in JuliaSyntax, and the APIs are still under-developed (`SyntaxNode` and `SourceFile` at least). Glad (not surprised!) to see you figured it out :)

The highlighting you're doing here looks very similar to the way diagnostics are show in JuliaSyntax and it looks like you might have rewritten some of the same tools?  Check out how diagnostics are printed at https://github.com/JuliaLang/JuliaSyntax.jl/blob/e57b80044e3a863924909ef51e1370c381109e58/src/diagnostics.jl#L44

We should share tools for pretty printing upstream in JuliaSyntax :heart:

src/sourcetext.jl
```diff
@@ -0,0 +1,205 @@
...
+function show_annotated(io::IO, src::CodeInfo, lineno::Int, @nospecialize(rt), mi::MethodInstance, rootnode::SyntaxNode;
+                        iswarn::Bool=false, hide_type_stable::Bool=false)
+    # FIXME: `lineno` will be useful for handling edited (`Revise`d) files
+    # Function signature
+    hd = untokenize(head(rootnode))
+    while hd != "=" && hd != "function"
+        hd == "macrocall" || error("expected function definition, got head ", hd)
+        rootnode = rootnode.val[end]
+        hd = untokenize(head(rootnode))
+    end
```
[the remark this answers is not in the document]
c42f: > Like `rootnode[end]` should maybe work

https://github.com/JuliaLang/JuliaSyntax.jl/pull/189

src/sourcetext.jl
```diff
@@ -0,0 +1,205 @@
...
+    clidx = find_codeloc(src, lineno)
+    ibegin = searchsortedfirst(src.codelocs, clidx)
+    ibegin += src.codelocs[ibegin] > lineno
+    iend = searchsortedlast(src.codelocs, clidx)
+    return ibegin:iend
+end
+
+function match_call(callname, src, lineno, taken)
+    @assert axes(taken) == axes(src.code)
+    clidx = find_codeloc(src, lineno)
```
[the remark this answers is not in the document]
timholy: I didn't annotate it, but here `src::Core.CodeInfo`. These are to find the corresponding lines in the type-inferred code. Briefly, `src.linetable[src.codelocs[i]]` is the `LineInfoNode`  corresponding to the type-inferred expression `src.code[i]`.

timholy:
Thanks @c42f! Your timing is excellent; just this morning I was thinking that I needed to rework all the printing here. The idea of upstreaming anything possible to JuliaSyntax sounds like the right design.

In some of my local work (not yet pushed because it's a little too ugly still), I've started on the first point in my [TODO list](https://github.com/JuliaDebug/Cthulhu.jl/pull/345#issuecomment-1422598455). Here's a screenshot of the current output:

image

As you can see, I've started adding some warnings that some calls couldn't be found in the type-inferred code. The printing of the line number is unlikely to be permanent, it's mostly for me to check whether I've gotten the line number right. (In case you're curious, for the cases here, some of them can't be found because type-inference determines that they will never run (given the input types the condition is false), others because the code is actually split out into an anonymous function (which is a separate `CodeInfo`), etc.) I added these warnings because there are going to be cases where we can't map a source expression to a type-inferred result, and it would be concerning if we thus missed the "true source" of a type-instability and users wasted time trying to fix a problem in the wrong location. I think these warnings at least ensure that we're confessing that there's some stuff we aren't fully reporting to them.

So one issue is that I'll need not only to print type-annotations after some `SyntaxNode`s but also detect when I've reached the end of the line, and then do some additional printing. Thus the printing tools in JuliaSyntax might need to allow some kind of check on whether you've encountered a `\n` while printing text, and if so interrupt with some different output. A character-stream iterator that would leave this code "in charge" of the output would accomplish that very nicely, I think.

src/sourcetext.jl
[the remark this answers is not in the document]
c42f: Ah fair enough!  I skimmed quickly and was thinking "this looks almost like the names of functions I've already written" haha :)

c42f:
> warnings that some calls couldn't be found in the type-inferred code

Cool. Interspersing diagnostics is tricky. I so want to preserve the "shape of the original code" in pretty printing. But in plain text that conflicts with providing this extra diagnostic info.  We discussed this back and forth a little in https://github.com/JuliaLang/JuliaSyntax.jl/issues/150

I can't really see a way around some kind of annotations system which will break the shape of the original code.  I had some ideas for fancier browser-based UI in https://github.com/JuliaLang/JuliaSyntax.jl/issues/150 (still somewhat integrated with the terminal experience via ANSI escape code hyperlinks) but we'll still need something which works well in a plain terminal by itself.

timholy:
Thanks for the many improvements, folks!

timholy:
OK, I think I've belatedly come up with a much better design. The idea is to orthogonalize the analysis (the matching between text & types) and the printing:
- create a variant of `SyntaxNode` that stores the inferred type of the return value of the node: this gives us the analog of type-inferred code but for source-text. Advantage: many different "consumers" could take advantage of the result and do whatever they want with it. 
- define printing for such type-annotated SyntaxNodes primarily in JuliaSyntax or extensible functions with methods defined in a different package. Advantage: it becomes a minor extension of JuliaSyntax and does not need any back-and-forth.

Note, the "unmached calls" could in fact be handled by defining a special type `Unmatched` and assigning it for any type-annotation where the match was uncertain. We could, e.g., print it as a yellow `::U` to indicate unmatched; it would be shorter and more informative than what I display above.

I welcome additional suggestions, so please share any that occur. But to me this seems obviously better than the approach I've taken here, so I'll probably get started on this tomorrow morning.

timholy:
I think I am ready to move forward with a more "final" implementation. I think the key question is how to organize the codebases. @vchuravy, have you decided whether you'd prefer the UI part to be separate from Cthulhu?

If it should be separate, then I propose to create a package ("TypeBrowser"?) in the JuliaDebug org that would have the Cthulhu-like "descending" UI but oriented towards source text; this repo could also have a subdir package ("TypedSyntax"?) that handles the matching between the trees produced by JuliaSyntax and type-inferred code. The output of "TypedSyntax" would be a [`TreeNode{TypedSyntaxData}`](https://github.com/JuliaLang/JuliaSyntax.jl/pull/193) and thus usable in other applications without loading the UI components.

If instead you want it to live under the Cthulhu umbrella, then I'd propose the "TypedSyntax" module could either be a standalone package (JuliaDebug? or other org?) or could live in this repo as a subdir package. The UI elements would be added to Cthulhu itself.

vchuravy:
Let's keep it with Cthulhu.

timholy:
Still very much of a WIP, but you can see hints of the final shape at https://github.com/JuliaDebug/Cthulhu.jl/blob/teh/sourcetext/TypedSyntax/README.md

I have some local changes to JuliaSyntax that I still need to split out into PRs, so don't expect it to work for you. But it does for me! 🙂

TypedSyntax/src/show.jl
```diff
@@ -0,0 +1,201 @@
...
+            print(io, "::NF")
+        end
+        return
+    end
+    if iswarn
+        color = !is_type_unstable(T) ? :cyan :
+                 is_small_union(T) ? :yellow : :red
+        printstyled(io, "::", T; color)
+    else
+        print(io, "::", T)
```
simeonschaub, line 171: Should we still print this in cyan or perhaps just bold? Otherwise these might get confused for manual type annotations.

simeonschaub (commented):
This is so cool!

Re how to ship this: I was wondering before whether we might want to have some third-party plugin mechanism for Cthulhu where external packages can hook in to display additional information alongside IR or add modes like this. That's probably a bigger change though, so I think for now, shipping this with Chulhu is fine.

c42f:
This is looking amazing!

Reading the README, it struck me that `SyntaxNode`'s printing of both line/column and byte range is a bit superfluous at this point. Perhaps I should remove the byte range by default.

timholy:
Formerly, this blindly matched calls based only on the function name. I've since dramatically restructured to use more of a data-flow approach: it progresses from all the symbols in the code and looks for expressions that contain the same args as a `:call/:invoke` statement in the inferred code. This seems to be much safer than resolving ambiguities based on order, as there turn out to be many examples where lowering alters the order of statements. The README probably needs some updating because I may have fixed some of the "caveats" I mentioned for the first iteration.

@aviatesk, it's possible that during the refactor I broke some of the things you added. If you test it and find some things are broken, please feel free to simply provide examples of what doesn't work, and I'll fix it.

I am guessing this is starting to get fairly decent. Unless more concerns arise in my explorations, I will next start working on the UI part in Cthulhu itself.

TypedSyntax/src/node.jl
```diff
@@ -0,0 +1,339 @@
...
+        for c in (kind(node) == K"do" ? (child(node, 1),) : children(node))  # process only `g(args...)` in `g(args...) do ... end`
+            collect_symbol_nodes!(symlocs, c)
+        end
+    end
+    return symlocs
+end
+
+# Find all places in the source code where a symbol is used
+function collect_symbol_nodes(rootnode)
+    kind(rootnode) ∈ KSet"function =" || error("expected function definition, got ", sourcetext(kind(rootnode)))
```
aviatesk, line 167: I think we need to register new version of JuliaSyntax so that TypedSyntax can use it.
After `dev`ing JuliaSyntax to the latest master, I tried `TypedSyntaxNode` for my test case `TypedSyntaxNode(sin, (Int,))` and I got error here. It looks like `rootnode` can be `@eval`ed function definition and so we want to handle it here somehow?

timholy: A new release would be good. I also forgot I had a couple more local modifications, you may need https://github.com/JuliaLang/JuliaSyntax.jl/pull/208. But your example was useful, fixed in the latest commit.

timholy: BTW, you've probably noticed this already, but I deliberately placed a bunch of `@assert`s in lots of places. The new design does involve understanding the code at a deeper level, and the `@assert`s are intended to make it fail as soon as something unexpected arises. It's going to be very fragile in these early days, but deliberately so. Feel free to just list test cases that fail and I will address them.

timholy:
Things are starting to semi-work. E.g, 
```julia
@descend optimize=false iswarn=true annotate_source=true hide_type_stable=true mapslices(sum, A; dims=2)
```
and then choose the `#mapslices#193` (or however it's numbered on your version of Julia) call.

aviatesk:
I think this is getting solid. How about colors types with `:cyan` or `:light_black` by default?

And here are some more cases that can be improved:
```julia
julia> function foo(x)
           x &= x
       end

julia> descend(foo, (Real,); optimize=false, annotate_source=true)
foo(x) @ Main REPL[19]:1
function foo(x::Real)::Any
    x &= x
end
Toggles: [o]ptimize, [w]arn, [h]ide type-stable statements, [d]ebuginfo, [r]emarks, [e]ffects, [i]nlining costs, [t]ype annotations, [s]yntax highlight for Source/LLVM/Native.
 • ↩

julia> function bar(x)
           x -= x
           return x
       end
bar (generic function with 1 method)

julia> descend(bar, (Real,); optimize=false, annotate_source=true)
bar(x) @ Main REPL[21]:1
function bar(x::Real)::Any
    x -= x
    return x
end
Toggles: [o]ptimize, [w]arn, [h]ide type-stable statements, [d]ebuginfo, [r]emarks, [e]ffects, [i]nlining costs, [t]ype annotations, [s]yntax highlight for Source/LLVM/Native.
 • ↩
```

timholy:
Those were very useful test cases, thanks!

timholy:
image

Final push! Must-haves:
- [ ] source-text in the sub-menu
- [ ] implement toggling between source- and typed-views
- [ ] tests of the UI in source-view
- [ ] change Cthulhu default to source-view
- [ ] documentation (usage, setting default config preferences)

Nice-to-haves (can be implemented as patch releases):
- [ ] see if more callsites can be mapped to source text
- [ ] fix line-number irregularities
- [ ] type-annotations from literals

Ideally I'd love to merge this by Monday, but if that seems rushed we can wait (I can tell my class to check out a branch).

timholy:
Documentation in the new README, see https://github.com/JuliaDebug/Cthulhu.jl/blob/teh/sourcetext/README.md

timholy:
This breaks some tests locally, I am still investigating.

But regardless, tests will fail until we register TypedSyntax. What version number should we start with? I'm a bit less of a fan of `0.x` version numbers than I once was, so I'd be inclined to start with `1.0.0` even though this is still fairly "beta." But alternatively we can use `0.1` (I'm just not sure when we'd switch to `1.0`.)

timholy:
Locally all tests pass; the only failure (on 1.9) is the fact that TypedSyntax can't be registered until after this merges. The tests of the new UI mode are minimal but probably good enough to merge. But I will hold off in case someone wants to review this.

aviatesk:
Some issues I found during playing with the latest commit:

- How about falling back to the previous lowered code view when we can't retrieve a source code?
  For example, I often use anonymous function as an entry, but it now fails since it's hard for `CodeTracking` to find a source of method created by `do` block:
  ```julia
  julia> descend((Int,)) do x
             [x]
         end
  ┌ Warning: couldn't retrieve source of (::var"#7#8")(x) @ Main REPL[14]:2
  └ @ Cthulhu ~/julia/packages/Cthulhu/src/reflection.jl:344
  ERROR: MethodError: no method matching getindex(::Nothing, ::Int64)
  Stacktrace:
   [1] find_callsites(interp::Cthulhu.CthulhuInterpreter, CI::Core.CodeInfo, stmt_infos::Vector{Core.Compiler.CallInfo}, mi::Core.MethodInstance, slottypes::Vector{Any}, optimize::Bool, annotate_source::Bool)
     @ Cthulhu ~/julia/packages/Cthulhu/src/reflection.jl:98
   [2] _descend(term::REPL.Terminals.TTYTerminal, interp::Cthulhu.CthulhuInterpreter, curs::Cthulhu.CthulhuCursor; override::Nothing, debuginfo::Symbol, optimize::Bool, interruptexc::Bool, iswarn::Bool, hide_type_stable::Bool, verbose::Nothing, remarks::Bool, with_effects::Bool, inline_cost::Bool, type_annotations::Bool, annotate_source::Bool)
     @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:496
   [3] _descend(term::REPL.Terminals.TTYTerminal, interp::Cthulhu.CthulhuInterpreter, mi::Core.MethodInstance; kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:iswarn,), Tuple{Bool}}})
     @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:760
   [4] _descend(term::REPL.Terminals.TTYTerminal, args::Any; interp::Core.Compiler.NativeInterpreter, kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:iswarn,), Tuple{Bool}}})
     @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:776
   [5] __descend_with_error_handling(args::Any; terminal::Any, kwargs::Base.Pairs{Symbol, V, Tuple{Vararg{Symbol, N}}, NamedTuple{names, T}} where {V, N, names, T<:Tuple{Vararg{Any, N}}})
     @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:239
   [6] _descend_with_error_handling(f::Any, argtypes::Any; kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:iswarn,), Tuple{Bool}}})
     @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:228
   [7] descend_code_typed(::Any, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
     @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:173
   [8] descend_code_typed(::Any, ::Any)
     @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:173
   [9] top-level scope
     @ REPL[14]:1
    ```

- And there is a bug with descending into `[x]`?
  ```julia
  julia> foo(x) = [x]

  julia> descend(foo, (Int,))
  foo(x) @ Main REPL[18]:1
  1 foo(x::Int64)::Vector{Int64} = [x::Int64]
  Toggles: [w]arn, [h]ide type-stable statements, [t]ype annotations, [s]yntax highlight for Source/LLVM/Native.
   • 1 [x]
     ↩
  ERROR: KeyError: key Symbol("") not found
  Stacktrace:
    [1] getindex
      @ ./dict.jl:485 [inlined]
    [2] map_ssas_to_source(src::Core.CodeInfo, rootnode::JuliaSyntax.TreeNode{JuliaSyntax.SyntaxData}, Δline::Int32)
      @ TypedSyntax ~/julia/packages/Cthulhu/TypedSyntax/src/node.jl:405
    [3] get_typed_sourcetext(mi::Core.MethodInstance, src::Core.CodeInfo, rt::Nothing)
      @ Cthulhu ~/julia/packages/Cthulhu/src/reflection.jl:350
    [4] find_callsites(interp::Cthulhu.CthulhuInterpreter, CI::Core.CodeInfo, stmt_infos::Vector{Core.Compiler.CallInfo}, mi::Core.MethodInstance, slottypes::Vector{Any}, optimize::Bool, annotate_source::Bool)
      @ Cthulhu ~/julia/packages/Cthulhu/src/reflection.jl:28
    [5] _descend(term::REPL.Terminals.TTYTerminal, interp::Cthulhu.CthulhuInterpreter, curs::Cthulhu.CthulhuCursor; override::Nothing, debuginfo::Cthulhu.DInfo.DebugInfo, optimize::Bool, interruptexc::Bool, iswarn::Bool, hide_type_stable::Bool, verbose::Nothing, remarks::Bool, with_effects::Bool, inline_cost::Bool, type_annotations::Bool, annotate_source::Bool)
      @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:496
    [6] _descend(term::REPL.Terminals.TTYTerminal, interp::Cthulhu.CthulhuInterpreter, curs::Cthulhu.CthulhuCursor; override::Nothing, debuginfo::Symbol, optimize::Bool, interruptexc::Bool, iswarn::Bool, hide_type_stable::Bool, verbose::Nothing, remarks::Bool, with_effects::Bool, inline_cost::Bool, type_annotations::Bool, annotate_source::Bool)
      @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:624
    [7] _descend(term::REPL.Terminals.TTYTerminal, interp::Cthulhu.CthulhuInterpreter, mi::Core.MethodInstance; kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:iswarn,), Tuple{Bool}}})
      @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:760
    [8] _descend(term::REPL.Terminals.TTYTerminal, args::Any; interp::Core.Compiler.NativeInterpreter, kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:iswarn,), Tuple{Bool}}})
      @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:776
    [9] __descend_with_error_handling(args::Any; terminal::Any, kwargs::Base.Pairs{Symbol, V, Tuple{Vararg{Symbol, N}}, NamedTuple{names, T}} where {V, N, names, T<:Tuple{Vararg{Any, N}}})
      @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:239
   [10] _descend_with_error_handling(f::Any, argtypes::Any; kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:iswarn,), Tuple{Bool}}})
      @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:228
   [11] descend_code_typed(::Any, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:173
   [12] descend_code_typed(::Any, ::Any)
      @ Cthulhu ~/julia/packages/Cthulhu/src/Cthulhu.jl:173
   [13] top-level scope
      @ REPL[19]:1
  ```

TypedSyntax/README.md
```diff
@@ -0,0 +1,156 @@
...
+This package aims to map types, as determined via type-inference, back to the source code as written by the developer. It can be used to understand program behavior and identify causes of "type instability" (inference failures) without the need to read [intermediate representations](https://docs.julialang.org/en/v1/devdocs/ast/) of Julia code.
+
+This package is built on [JuliaSyntax](https://github.com/JuliaLang/JuliaSyntax.jl) and extends it by attaching type annotations to the nodes of its syntax trees. Here's a demo:
+
+```julia
+julia> using TypedSyntax
+
+julia> f(x, y, z) = x + y * z;
+
+julia> node = TypedSyntaxNode(f, (Float64, Int, Float32))
```
aviatesk, line 12:
```suggestion
```

TypedSyntax/README.md
```diff
@@ -0,0 +1,156 @@
...
+│   %13 = Base.not_int(%12)::Bool
+└──       goto #4 if not %13
+3 ─       goto #2
+    @ REPL[3]:6 within `summer`
+4 ┄       return s
+) => Union{Float64, Int64}
+```
+and very few calls here map to the source:
+```julia
+16-element Vector{Vector{Union{JuliaSyntax.TreeNode{JuliaSyntax.SyntaxData}, JuliaSyntax.TreeNode{TypedSyntax.TypedSyntaxData}}}}:
```
aviatesk, line 112: How can we get this view?

TypedSyntax/README.md
```diff
@@ -0,0 +1,156 @@
...
+Second, this package works by attempting to "reconstruct history": starting from the type-inferred code, it tries to map calls back to the source. It would be much safer to instead keep track of the source during inference, but at present this is not possible (see [this Julia issue](https://github.com/JuliaLang/julia/issues/31162)). There are cases where this mapping fails: for example, with
+
+```julia
+julia> function summer(list)
+           s = 0
+           for x in list
+               s += x
+           end
+           return s
+       end;
```
aviatesk, line 82: It seems like we need to remove this semicolon otherwise we can't construct `TypedSyntaxNode`:
```julia
julia> function summer(list)
           s = 0
           for x in list
               s += x
           end
           return s
       end;

julia> TypedSyntaxNode(summer, (Vector{Float64},))
ERROR: MethodError: no method matching iterate(::Nothing)

Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen})
   @ Base range.jl:887
  iterate(::Union{LinRange, StepRangeLen}, ::Integer)
   @ Base range.jl:887
  iterate(::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}}
   @ Base dict.jl:716
  ...

Stacktrace:
 [1] indexed_iterate(I::Nothing, i::Int64)
   @ Base ./tuple.jl:93
 [2] JuliaSyntax.TreeNode{TypedSyntax.TypedSyntaxData}(f::Any, t::Any; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ TypedSyntax ~/julia/packages/Cthulhu/TypedSyntax/src/node.jl:20
 [3] JuliaSyntax.TreeNode{TypedSyntax.TypedSyntaxData}(f::Any, t::Any)
   @ TypedSyntax ~/julia/packages/Cthulhu/TypedSyntax/src/node.jl:18
 [4] top-level scope
   @ REPL[33]:1
```

TypedSyntax/README.md
aviatesk, line 112: How can we get this view?

TypedSyntax/src/show.jl
```diff
@@ -0,0 +1,162 @@
...
+function countconcrete(@nospecialize(T))
+    if Base.isdispatchelem(T)
+        return 1, true
+    elseif isa(T, Union)
+        na, isca = countconcrete(T.a)
+        nb, iscb = countconcrete(T.b)
+        return na + nb, isca & iscb
+    end
+    return 0, false
+end
```
aviatesk, line 162:
```suggestion
end

```

src/Cthulhu.jl
```diff
@@ -416,7 +421,8 @@ function _descend(term::AbstractTerminal, interp::AbstractInterpreter, curs::Abs
     remarks::Bool                            = CONFIG.remarks&!CONFIG.optimize,      # default is false
     with_effects::Bool                       = CONFIG.with_effects,                  # default is false
     inline_cost::Bool                        = CONFIG.inline_cost&CONFIG.optimize,   # default is false
-    type_annotations::Bool                   = CONFIG.type_annotations               # default is true
+    type_annotations::Bool                   = CONFIG.type_annotations,              # default is true
+    annotate_source::Bool                    = CONFIG.annotate_source,               # default is false
```
aviatesk, line 425:
```suggestion
    annotate_source::Bool                    = CONFIG.annotate_source,               # default is true
```

TypedSyntax/README.md
timholy: It required some internals. I refactored to make it easier to obtain and modified the README accordingly.

timholy (commented):
Thanks for the review!

I fixed the issues you noticed. It now falls back to the typed view if the source can't be found. (That's a very important change, thanks!)

timholy:
One thought I had: should I split this into two PRs?

- one PR is to add `TypedSyntax` (assuming we still want it to be a subdir package here, or we can move it to a separate repo). We could register it before moving on to the next PR.
- another PR is to wire it into Cthulhu and make this the default mode.

aviatesk:
> One thought I had: should I split this into two PRs?
> 
> * one PR is to add `TypedSyntax` (assuming we still want it to be a subdir package here, or we can move it to a separate repo). We could register it before moving on to the next PR.
> * another PR is to wire it into Cthulhu and make this the default mode.

I agree with it :)

timholy:
See #353

PR #348, opened 2023-02-10
fix tests on nightly

aviatesk:
Still there are remaining failures due to code coverage. 
It often confuses our CI and is not that important. I will disable it entirely if possible.


aviatesk:
Okay, now I disabled `code-coverage` and `check-bounds=yes` options in CI so that our CI runs in the same enviroment as we use locally with the default command line options.
I'd say the coverage isn't worth to be turned on for this package, but please let me know if you want to keep code coverage on.

PR #353, opened 2023-02-28
Add TypedSyntax package

timholy:
This package attempts to map statements in type-inferred code back to the source code as written by the programmer.

The intention is to use this in Cthulhu to present the results of inference in an easier-to-digest form. There are, of course, potential additional applications of this source-mapping, which is why it is developed as a semi-independent package.

This is part of #345, split out so it can be registered before making Cthulhu dependent on it.

### DIFF
--- a/TypedSyntax/README.md
+++ b/TypedSyntax/README.md
@@ -1,0 +1,133 @@
+# TypedSyntax
+
+This package aims to map types, as determined via type-inference, back to the source code as written by the developer. It can be used to understand program behavior and identify causes of "type instability" (inference failures) without the need to read [intermediate representations](https://docs.julialang.org/en/v1/devdocs/ast/) of Julia code.
+
+This package is built on [JuliaSyntax](https://github.com/JuliaLang/JuliaSyntax.jl) and extends it by attaching type annotations to the nodes of its syntax trees. Here's a demo:
+
+```julia
+julia> using TypedSyntax
+
+julia> f(x, y, z) = x + y * z;
+
+julia> node = TypedSyntaxNode(f, (Float64, Int, Float32))
+line:col│ tree                                   │ type
+   1:1  │[=]                                     │Float64
+   1:1  │  [call]
+   1:1  │    f
+   1:3  │    x                                   │Float64
+   1:6  │    y                                   │Int64
+   1:9  │    z                                   │Float32
+   1:13 │  [call-i]                              │Float64
+   1:14 │    x                                   │Float64
+   1:16 │    +
+   1:17 │    [call-i]                            │Float32
+   1:18 │      y                                 │Int64
+   1:20 │      *
+   1:22 │      z                                 │Float32
+```
+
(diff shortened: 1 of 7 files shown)


timholy:
Nice, one can run the TypedSyntax tests as part of CI even before Cthulhu knows about this package. Makes sense, but still nice.

TypedSyntax/src/node.jl
```diff
@@ -0,0 +1,432 @@
...
+
+# function sparam_name(mi::MethodInstance, i::Int)
+#     sig = (mi.def::Method).sig::UnionAll
+#     while true
+#         i == 1 && break
+#         sig = sig.body::UnionAll
+#         i -= 1
+#     end
+#     return sig.var.name
+# end
```
timholy, line 160: @aviatesk, I didn't find a need to wire this in with the new design, but if you know of one, please provide a (failing) test case!

TypedSyntax/src/show.jl
```diff
@@ -0,0 +1,165 @@
...
+        for n in children(node)
+            JuliaSyntax._show_syntax_node(io, current_filename, n, new_indent, show_byte_offsets)
+        end
+    end
+end
+
+
+function Base.printstyled(io::IO, rootnode::MaybeTypedSyntaxNode;
+                          type_annotations::Bool=true, iswarn::Bool=true, hide_type_stable::Bool=true,
+                          idxend = last_byte(rootnode), kwargs...)
```
aviatesk, line 32:
```suggestion
                          idxend = last_byte(rootnode))
```
Since `kwargs` is unused at all, I think we should remove it since otherwise we will accept whatever keyword argument silently.

TypedSyntax/README.md
```diff
@@ -0,0 +1,133 @@
...
+julia> function summer(list)
+           s = 0
+           for x in list
+               s += x
+           end
+           return s
+       end;
+```
+then (on Julia 1.9)
+```
```
aviatesk, line 84:
```suggestion
```julia
```

TypedSyntax/src/node.jl
timholy: I'd argue for `(myzero(x::Float64::T) where T)::Float64` or `(myzero((x::Float64)::T) where T)::Float64`; I think "binding" the `Float64` more tightly to `x` than the type restriction makes sense; after all `Float64` is the inferred type of `x`, and printing the type in the inner-most operation makes sense. This choice treats `::` basically as any other infix operator.

aviatesk: Okay. Now I think `(myzero((x::Float64)::T) where T)::Float64` is the most reasonable option but I would like to leave the decision to you.

aviatesk (approved):

timholy:
I added `Expr(:static_parameter, i)` support, thanks for contributing it.

It's starting to feel decent, though future changes are a certainty. Unless there are objections, I'll merge and register later today.

PR #354, opened 2023-03-01
Add RTCallInfo for runtime calls

timholy:
The list of call sites serves two purposes:

- it gives you options to descend into
- it summarizes the calls made by your function

Given the second purpose, I've sometimes been confused by the absence of an entry for the "worst" of all cases, runtime dispatch. This seems particularly accute for #345, but may have value even when examining the CodeInfo/IRCode.

Here's a demo: on `master`,
```julia
julia> function summer(list)
           s = 0      # deliberately not `zero(eltype(list))`
           for x in list
               s += x
           end
           return s
       end;

julia> @descend optimize=false summer([1.0, 2.0])
 ⋮
 • %3 = iterate(::Vector{Float64})::Union{Nothing, Tuple{Float64, Int64}}
   %10 = call → +(::Union{Float64, Int64},::Float64)::Float64
   %11 = iterate(::Vector{Float64},::Int64)::Union{Nothing, Tuple{Float64, Int64}}
   ↩
```
but
```julia
julia> @descend optimize=false summer(Any[1.0, 2.0])
 ⋮
 • %3 = iterate(::Vector{Any})::Union{Nothing, Tuple{Any, Int64}}
   %11 = iterate(::Vector{Any},::Int64)::Union{Nothing, Tuple{Any, Int64}}
   ↩
```
The absence of any "acknowledgement" of the call to `+` gave me pause.

In contrast, on this branch the first case is unchanged, whereas the second is
```julia
 • %3 = iterate(::Vector{Any})::Union{Nothing, Tuple{Any, Int64}}
   %10 = RT call +(::Any,::Any)::Any
   %11 = iterate(::Vector{Any},::Int64)::Union{Nothing, Tuple{Any, Int64}}
   ↩
```

I've set it up so that attempting to descend into the RT call gives this:
```
   %3 = iterate(::Vector{Any})::Union{Nothing, Tuple{Any, Int64}}
 • %10 = RT call +(::Any,::Any)::Any
   %11 = iterate(::Vector{Any},::Int64)::Union{Nothing, Tuple{Any, Int64}}
   ↩
[ Info: This is a runtime call. You cannot descend into it.
Toggles: [o]ptimize, [w]arn, [h]ide type-stable statements, [d]ebuginfo, [r]emarks, [e]ffects, [i]nlining costs, [t]ype annotations, [s]yntax highlight for Source/LLVM/Native.
 • %3 = iterate(::Vector{Any})::Union{Nothing, Tuple{Any, Int64}}
   %10 = RT call +(::Any,::Any)::Any
   %11 = iterate(::Vector{Any},::Int64)::Union{Nothing, Tuple{Any, Int64}}
   ↩
```

I *think* filtering out all the `Core.Builtin` calls will prevent this from being too "noisy," but I'd be grateful for the thoughts of those who know much more about inference than me.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -21,7 +21,7 @@ FoldingTrees = "1"
 JuliaSyntax = "0.3.2"
 Preferences = "1"
 SnoopPrecompile = "1"
-TypedSyntax = "1.0.1"
+TypedSyntax = "1.0.2"
 julia = "1.7"
 
 [extras]
--- a/README.md
+++ b/README.md
@@ -78,8 +78,17 @@ In the final section, you see:
 ![calls](images_readme/descend_calls.png)
 
 This is a menu of calls that you can further descend into. Move the dot `•` with the up and down
-arrow keys, and hit Enter to descend into a particular call. Calls that start with `%nn = ...`
-are in Julia's internal [Abstract Syntax Tree (AST)](https://docs.julialang.org/en/v1/devdocs/ast/) form;
+arrow keys, and hit Enter to descend into a particular call. Any calls that are made at runtime ([dynamic dispatch](https://en.wikipedia.org/wiki/Dynamic_dispatch)) cannot be descended into;
+if you select one, you'll see
+
+```
+[ Info: This is a runtime call. You cannot descend into it.
+```
+
+and the call menu will be printed again.
+
+Calls that start with `%nn = ...` are in Julia's internal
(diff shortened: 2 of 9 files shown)


timholy:
I'm planning to merge this tomorrow, too.

vchuravy (approved):

vchuravy:
Maybe print it as `runtime <+(::Any,::Any)::Any>` or `dynamic` It might not be clear what RT stands for.

PR #356, opened 2023-03-02
Implement annotated source-view & make default

timholy:
This leverages TypedSyntax to attach type-annotations to user source code (see the [README on this branch](https://github.com/JuliaDebug/Cthulhu.jl/blob/teh/sourcetext2/README.md) for an overview). This is intended as a more "newbie-friendly" mode for Cthulhu. Because source text is typically less verbose than IRCode, it may even be useful in some circumstances for Cthulhu veterans.

This is the second half of #345. Tests pass locally but will not pass here until https://github.com/JuliaRegistries/General/pull/78815 merges. But I thought it would be better to post this now to give folks time to review these changes.

Closes #345.


timholy:
TypedSyntax is now registered. Let's re-run the tests.

timholy:
You've already given this a spin, I know, and I appreciate it! And implementation details can always be fixed later, so if you're good with it then I think that makes sense.

For anyone who hasn't tested this yet and who may understandably be a bit worried about it, perhaps the best option is to just try the branch out with, e.g., the demo in the [README](https://github.com/JuliaDebug/Cthulhu.jl/blob/teh/sourcetext2/README.md). That will let you get a better understanding of what the user experience is like and whether it's smooth to switch back to traditional IRCode/CodeInfo views. I won't merge until I've played with that again myself, but any other eyeballs (and especially, real-world usage, which I've done very little of so far) are always appreciated.

timholy:
Whew, no 1.7 failures.

timholy:
I've done much more real-world usage of this and handled a wider variety of code. Unless objections arise, I'll merge & register tomorrow. I *think* Cthulhu 2.9.0 is a good version number for this, but if anyone would prefer to go to 3.0.0, let me know.

PR #319, opened 2022-08-30
Support semi concrete eval

ianatol:
Rebase of #280

PR #357, opened 2023-03-05
Add Julia 1.7 to test matrix

timholy:
Project.toml claims we support 1.7; given how closely this package
follows Julia development, that's a dangerous claim if it's not tested.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,6 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1.7'
           - '1.8'
           - '~1.9.0-0'
           - 'nightly'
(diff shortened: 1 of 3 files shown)


vchuravy (approved):

timholy:
As I feared, on Julia 1.7 there's no `@assume_effects`. This might be trivially fixable: it only appears in tests. If Julia 1.7 is genuinely broken and can't easily be fixed (here, not in Julia), someone needs to run `RetroCap.jl` to General.

e0789fa7 changed the CI line from '1' to '1.8'; that might mean it was broken before that. I'm guessing it was broken in 6a5a21bc (#319, CC @ianatol), but obviously only because CI.yml was not explicit about what '1' meant.

timholy:
OK, I just skipped over the failing tests. To me this seems OK, so I'll go ahead and merge and then re-run #356.

PR #361, opened 2023-03-08
Add link to Youtube video for source-view

timholy:
I don't want to overshadow @vchuravy's video, but it seems like it might be time for an updated demo. If you don't like this, feel free to close! Or suggest alternate placement of the link and/or wording changes to the README.


aviatesk (approved):

aviatesk:
Let's merge this as we've not got any objections for two weeks.

PR #407, opened 2023-03-21
Clarify menu options for ascend

timholy:
From ProfileView, here's how the new menu looks:

image

Hopefully should be clearer what the options really are.


timholy:
Test failure on nightly is unrelated to this PR. (See https://github.com/JuliaLang/julia/pull/49071.)

PR #434, opened 2023-04-18
Use compileable_specialization internal API

vtjnash:
The code doesn't seem to care about `:compilesig`, and it not meaningful to inference either, as it turns out (https://github.com/JuliaLang/julia/pull/49404)


vtjnash:
merge me?

Issue #443, opened 2023-04-27
`no method matching children(::Cthulhu.Callsite)` while descending into `indexed_iterate`


jishnub:
```julia
julia> @descend (A -> ((λ,v) = eigen(A); (λ,v)))(rand(2,2))
(::var"#11#12")(A) @ Main REPL[8]:1
┌ Warning: couldn't retrieve source of (::var"#11#12")(A) @ Main REPL[8]:1
└ @ TypedSyntax ~/.julia/packages/TypedSyntax/AZkb2/src/node.jl:31
Variables
  #self#::Core.Const(var"#11#12"())
  A::Matrix{Float64}
  @_3::Val{:vectors}
  v::Union{Matrix{ComplexF64}, Matrix{Float64}}
  λ::Union{Vector{ComplexF64}, Vector{Float64}}

∘ ─ %0 = invoke #11(::Matrix{Float64})::Tuple{Union{Vector{ComplexF64}, Vector{Float64}}, Union{Matrix{ComplexF64}, Matrix{Float64}}}
    @ REPL[8]:1 within `#11`
1 ─ %1 = Main.eigen(A)::Union{Eigen{ComplexF64, ComplexF64, Matrix{ComplexF64}, Vector{ComplexF64}}, Eigen{Float64, Float64, Matrix{Float64}, Vector{Float64}}}
│   %2 = Base.indexed_iterate(%1, 1)::Union{Tuple{Vector{ComplexF64}, Val{:vectors}}, Tuple{Vector{Float64}, Val{:vectors}}}
│        (λ = Core.getfield(%2, 1))::Union{Vector{ComplexF64}, Vector{Float64}}
│        (@_3 = Core.getfield(%2, 2))::Core.Const(Val{:vectors}())
│   %5 = Base.indexed_iterate(%1, 2, @_3)::Union{Tuple{Matrix{ComplexF64}, Val{:done}}, Tuple{Matrix{Float64}, Val{:done}}}
│        (v = Core.getfield(%5, 1))::Union{Matrix{ComplexF64}, Matrix{Float64}}
│   %7 = Core.tuple(λ, v)::Tuple{Union{Vector{ComplexF64}, Vector{Float64}}, Union{Matrix{ComplexF64}, Matrix{Float64}}}
└──      return %7
Select a call to descend into or ↩ to ascend. [q]uit. [b]ookmark.
[...]

PR #444, opened 2023-04-27
Check type of sourcenode before calling `children`

timholy:
xref #443

This fixes the immediate error, but descending further triggers another one that we probably need @aviatesk to help debug.


timholy:
CC @jishnub

timholy:
Since this does make sense on its own, I'll merge it. @jishnub, perhaps you can file the error that remains after this is fixed?

PR #454, opened 2023-05-13
Improve inference

timholy:
These arose from an analysis with JET. A subset of the changes exist only to silence JET warnings, but the large majority of changes are productive changes that will improve performance and make it more resistant to invalidation.


aviatesk:
Great to see this PR!

timholy:
Was good get back to using JET. It's likely to be quite important in practice for static compilation, so I wanted to take it out for a spin. It's evolved very impressively!

PR #457, opened 2023-06-01
drop support for 1.7

aviatesk:
It would be good to remove the compat complexities in our code base.


timholy (approved):

vchuravy:
I am okay with dropping 1.7, but 1.8 is still used in the wild.

timholy:
But old versions still work there? Though perhaps we should merge and release #459 first.

vchuravy:
> But old versions still work there?

But there are still bugs on it :) and backporting seems not like something we want to sign up.
I rather trade a slight increase in code complexity for being able to get consistent answers when investigating behaviour across versions.

timholy:
What, code with 🐞? Surely you jest!

Can't we tell people to upgrade to Julia 1.9 if they want Cthulhu with the latest fixes? They should be doing that anyway (yes I'm biased).

But if that's not really viable, then I'd agree that we really don't want to have to get into backporting and it would be better to wait a bit longer before dropping 1.8.

vchuravy:
It's more me switching back between 1.8 and 1.9 to figure out what changed in the compiler lead to a different inference result.

aviatesk:
Which 1.8 version you want to support? I can try to maintain the compatibility for now then.

vchuravy:
1.8.5 is fine.

aviatesk:
This is just a ballpark idea, but how about we plan on dropping support for version 1.8 in a couple of months? Truthfully, we'd prefer to avoid exerting effort to maintain older versions as much as possible.

vchuravy:
I am okay with having a policy that's N-1, N. 
E.g. support the last two released versions of Julia.

PR #462, opened 2023-06-01
Fixes for `MethodInstance` objects that do not come from a Method

staticfloat:
When running `descend()` on a MethodInstance that does not come from a `Method`, we need to avoid attempting to tie back to the original source or method definition names.  This is the case when you have a chunk of IR that you have generated, but no real method to map back to it with.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -511,7 +511,7 @@ function _descend(term::AbstractTerminal, interp::AbstractInterpreter, curs::Abs
                     ioctx = IOContext(iostream,
                         :color => true,
                         :displaysize => displaysize(iostream), # displaysize doesn't propagate otherwise
-                        :SOURCE_SLOTNAMES => Base.sourceinfo_slotnames(codeinf),
+                        :SOURCE_SLOTNAMES => codeinf === nothing ? false : Base.sourceinfo_slotnames(codeinf),
                         :with_effects => with_effects)
                     stringify(ioctx) do lambda_io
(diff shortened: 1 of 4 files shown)


timholy (approved):

timholy:
Any chance of a test?

test/test_Cthulhu.jl
```diff
@@ -973,4 +973,28 @@ end
     end
 end
 
+@testset "Bare-bones MIs" begin
```
aviatesk, line 976:
```suggestion
@static VERSION ≥ v"1.9-" && @testset "Bare-bones MIs" begin
```

aviatesk (approved):

src/Cthulhu.jl
```diff
@@ -511,7 +511,7 @@ function _descend(term::AbstractTerminal, interp::AbstractInterpreter, curs::Abs
                     ioctx = IOContext(iostream,
                         :color => true,
                         :displaysize => displaysize(iostream), # displaysize doesn't propagate otherwise
-                        :SOURCE_SLOTNAMES => Base.sourceinfo_slotnames(codeinf),
+                        :SOURCE_SLOTNAMES => codeinf === nothing ? false : Base.sourceinfo_slotnames(codeinf),
```
aviatesk, line 514: Is this change necessary? `codeinf` here should always be `CodeInfo` object.

staticfloat: We did need it, yes, I'll send you a MWE.

PR #468, opened 2023-06-14
Print lifted values as Core.Const

### LINKED ISSUE #467: Should values as type parameters be displayed as `Const`s?
```julia
julia> function f(::Val{N}) where N
           if N == 2
               return 4
           else
               return 3
           end
       end

julia> @descend f(Val(4))
f(::Val{N}) where N @ Main REPL[1]:1
1 function (f(::(Val{N})::Val{4}) where N)::Core.Const(3)
2     if N::Type{4} == 2
3         return 4
4     else
5         return 3
6     end
7 end
Toggles: [w]arn, [h]ide type-stable statements, [t]ype annotations, [s]yntax highlight for Source/LLVM/Native.
 • %1 = < concrete eval > ==(::Core.Const(4),::Core.Const(2))::Core.Const(false)
   ↩
```
On line 2, we see `N::Type{4}`, but should this be `Core.Const(4)` instead?

vchuravy:
Fixes #467

### DIFF
--- a/TypedSyntax/src/node.jl
+++ b/TypedSyntax/src/node.jl
@@ -680,7 +680,13 @@ function map_ssas_to_source(src::CodeInfo, rootnode::SyntaxNode, Δline::Int)
                                 name = sparam_name(mi, id)
                                 for t in symlocs[name]
                                     T = mi.sparam_vals[id]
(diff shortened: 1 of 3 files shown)


timholy (approved):
Thanks!

PR #469, opened 2023-06-15
Integrate Cthulhu with VSCode - show types in source code

Zentrik:
Depends on https://github.com/julia-vscode/julia-vscode/pull/3328 for showing types inline, however can be merged in before.

This pr show types in the source code when using the repl in VSCode, e.g. `@descend f(1)` with warnings on gives
image
And without warnings 
image

There a couple improvements I still want to make to this pr: 

- [x] Adding a way to toggle the vscode integration (should this be implemented the same way the warn toggle works?)
- [x] Removing the duplication of tests in TypedSyntax (I tried putting the tests into a separate file and using `include` but then the `m = @which TSN.unnamedargs(Matrix{Float32}, Int; a="hello")` test started failing claiming `mi` was `nothing`) 
- [x] Adding a way to automatically show types in VSCode for any called functions in an open file without having to descend into them.

As an example of how to change the colour of the hints in VSCode, I'm using the Monokai theme with the following in my `settings.json`
```json
"workbench.colorCustomizations": {
        // Name of the theme you are currently using
        "[Monokai]": {
            // Overrides for specific kinds of inlay hints
            "editorInlayHint.typeForeground": "#66D9EF",
            "editorInlayHint.typeBackground": "#00000000",
            "editorInlayHint.parameterForeground": "#f8f8f2", // defaults from https://github.com/microsoft/vscode/blob/2e335d2df2001439b09483e7fcff8379ac1a7a16/extensions/theme-monokai/themes/monokai-color-theme.json#L41-L42
            "editorInlayHint.parameterBackground": "#75715E",
            // Colours for unstable types when warnings are on
            "editorInlayHint.foreground": "#e13030",
            "editorInlayHint.background": "#00000000",
        }
    }
```


timholy:
I haven't looked at the implementation yet, but I'm very excited to see this! I agree with the items on your TODO list that I understand. What's the precise issue about toggling the integration? Would that be something controlled from the VSCode GUI?

Zentrik:
I don't think there is any good way to toggle the integration from VSCode, the setting for the types and the warning diagnostics (the orange line saying Unstable Type) is Julia wide.
I was thinking the best way would be to add another toggle to `@descend` like the `[w]arn` one.

Zentrik:
On this pr it seems when descending most of the time (~75%) is spent on type inference and most of the remaining time is spent on `string(T)`, i.e. converting types to strings for pretty printing. I don't think there's an easy way to speed up either apart from memoizing the type to string conversions.

pfitzseb:
This is really cool!

I was playing around with this a little bit and have a few suggestions:
1. Auto-clear Cthulhu inlay hints when the `@descend` session ends
2. Auto-open the current function in the editor instead of printing the source code
3. Enable showing inlay hints/diagnostics by default

I'd also love it if the inserted parentheses would look different from the inlay hints, but afaict that's not currently supported by VS Code.

timholy:
I've gotten started reviewing this (sorry for the delay), but I think I need to ask for some clarification. What, exactly, is the issue you're trying to solve? If I run Cthulhu from the VSCode Julia REPL, I get this:

image

(Here I've hidden the type-stable statements.) What problem are you trying to fix?

Zentrik:
> This is really cool!
> 
> I was playing around with this a little bit and have a few suggestions:
> 
>     1. Auto-clear Cthulhu inlay hints when the `@descend` session ends
> 
>     2. Auto-open the current function in the editor instead of printing the source code
> 
>     3. ~Enable showing inlay hints/diagnostics by default~
> 
> 
> I'd also love it if the inserted parentheses would look different from the inlay hints, but afaict that's not currently supported by VS Code.

Thanks for the suggestions, I've implemented the first one. As Cthulhu already prints a hyperlink to the current function and because it seems like it would be annoying if the editor jumped everytime Cthulhu was used, for now I haven't implemented the second suggestion though I would be happy to if more people wanted it.

Zentrik:
> Is this to show the type-annotation in the editor pane?

Yes, this shouldn't change the repl output apart from adding some toggles. If you want to see the type-annotations in the editor pane you'll need to use the master branch of https://github.com/julia-vscode/julia-vscode.

> Can you point me to the requirements that VSCode has for doing this? (Have a link?)

Is this what you're asking for [https://code.visualstudio.com/api/references/vscode-api#InlayHint](https://code.visualstudio.com/api/references/vscode-api#InlayHint)?
This is the struct in which I'm storing the information needed for type annotations, https://github.com/JuliaDebug/Cthulhu.jl/blob/d123e959dcbffa4c1c40234855534bdcadea49c1/TypedSyntax/src/vscode.jl#L35-L40

Zentrik:
Here's a better screenshot 
image

timholy:
That is indeed nice. Is there the option of turning of regular syntax highlighting, so the only colors are from type-hinting?

One obvious point is that it's redundant with what appears in the REPL. I assume you still find this useful, perhaps particularly for long functions? We could also add, e.g., automatic hinting for every `Method` that has been specialized to a single `MethodInstance`.

Zentrik:
> Is there the option of turning of regular syntax highlighting, so the only colors are from type-hinting?

No, after looking into it very briefly it doesn't seem like there's an easy way to disable syntax highlighting.

> One obvious point is that it's redundant with what appears in the REPL. I assume you still find this useful, perhaps particularly for long functions?

As shown in the screenshot types are shown automatically for any functions called in `h` that are defined in the same file. This is probably the main use for showing types in VSCode.

> We could also add, e.g., automatic hinting for every `Method` that has been specialized to a single `MethodInstance`.

I'm not sure what it means for a `Method` to be specialized to a single `MethodInstance`, does this mean all arguments are concrete types at a given callsite?

timholy:
I just meant "it's only been compiled for a single concrete signature." Some PrecompileTools workloads do this, for example.

TypedSyntax/src/show.jl
```diff
@@ -43,19 +44,44 @@ function Base.printstyled(io::IO, rootnode::MaybeTypedSyntaxNode;
...
     end
-    position = show_src_expr(io, rootnode, position, "", ""; type_annotations, iswarn, hide_type_stable, nd)
+    position = show_src_expr(io, rootnode, position, "", "", type_hints, warn_diagnostics; type_annotations, iswarn, hide_type_stable, nd)
     catchup(io, rootnode, position, nd, idxend+1)   # finish the node
     return nothing
 end
-Base.printstyled(rootnode::MaybeTypedSyntaxNode; kwargs...) = printstyled(stdout, rootnode; kwargs...)
+function Base.printstyled(io::IO, rootnode::MaybeTypedSyntaxNode, 
+                          type_hints = Dict{String, Vector{InlayHint}}(), 
+                          warn_diagnostics = WarnUnstable[];
```
timholy, line 57: What's the reason for making some of these positional and the others keywords?

I think my main question would be: could we clean this up by adding these to the `io` via an `IOContext`? There appear to be good arguments for that, including that they are both outputs rather than inputs (it's a little weird to have the argument role go `(Output, Input, Output, Output)`), and `Base.printstyled` docstring specifies that any additional arguments get printed to `io` (and not *printed to*). It might also let this be less intrusive of a change.

TypedSyntax/src/vscode.jl
```diff
@@ -0,0 +1,80 @@
...
+        items = to_vscode_type.(warn_diagnostics),
+    )
+end
+function add_diagnostic!(warn_diagnostics, node, position, severity)
+    file_path = node.filename
+    line, column = source_location(node, position)
+    push!(warn_diagnostics, WarnUnstable(file_path, line, severity))
+end
+
+const InlayHintKinds = (
```
timholy, line 31: Should these be an `@enum`?

TypedSyntax/src/vscode.jl
```diff
@@ -0,0 +1,80 @@
...
+    filepath = node.filename
+    line, column = source_location(node, position)
+
+    if filepath ∉ keys(type_hints)
+        type_hints[filepath] = InlayHint[]
+    end
+    push!(type_hints[filepath], InlayHint(line-1, column, message, kind))
+end
+
+function show_annotation(io, @nospecialize(T), post, node, position, type_hints, warn_diagnostics; iswarn::Bool)
```
timholy, line 60: Here too I think getting `type_hints` and `warn_diagnostics` from `io` makes more sense, so you don't need very much vscode-specific code.

src/codeview.jl
```diff
@@ -121,9 +122,10 @@ cthulhu_typed(io::IO, debuginfo::DebugInfo, args...; kwargs...) =
     cthulhu_typed(io, Symbol(debuginfo), args...; kwargs...)
 function cthulhu_typed(io::IO, debuginfo::Symbol,
     src::Union{CodeInfo,IRCode}, @nospecialize(rt), effects::Effects, mi::Union{Nothing,MethodInstance};
-    iswarn::Bool=false, hide_type_stable::Bool=false,
+    iswarn::Bool=false, hide_type_stable::Bool=false, optimize::Bool=true,
```
timholy, line 125: Ooh, interesting, seems like this could be a separate PR (especially if you have a triggering issue that's independent of this PR). But I'm OK if it gets bundled here.

timholy (commented):
Thanks Zentrik and sorry again for the delay. This is clearly needed functionality, so I view this as on-track for merging. My one wish is to see if there are some cases where we can reduce the amount of code-duplication, while also improving the consistent of arguments, by bundling some of VSCode's needed outputs into an `IOContext(io, :type_hints => type_hints, :warn_diagnostics => warn_diagonstics)` and then "printing" to that object. The code that does the printing can try to extract these items in appropriate places with `type_hints = get(io, :type_hints, nothing)` and then branch on `type_hints === nothing`.

TypedSyntax/src/node.jl
```diff
@@ -35,7 +35,8 @@ function tsn_and_mappings(m::Method, src::CodeInfo, @nospecialize(rt); warn::Boo
 end
 
 function tsn_and_mappings(m::Method, src::CodeInfo, @nospecialize(rt), sourcetext::AbstractString, lineno::Integer; warn::Bool=true, strip_macros::Bool=false, kwargs...)
-    rootnode = JuliaSyntax.parsestmt(SyntaxNode, sourcetext; filename=string(m.file), first_line=lineno, kwargs...)
+    filename = isnothing(functionloc(m)[1]) ? string(m.file) : functionloc(m)[1]
```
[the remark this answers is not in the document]
Zentrik: For a function defined in the REPL `functionloc` will return `nothing` whilst `m.file` will give e.g. `Symbol("REPL[9]")`.

TypedSyntax/src/vscode.jl
Zentrik: As InlayHintKinds can also be `nothing` I've left it as a constant. I changed the severity in WarnUnstable (now Diagnostic) to an @enum, I put it in a module to avoid polluting the global namespace but I can change that.

src/codeview.jl
Zentrik: I'm pretty sure the only reason I'm passing in `optimize` is because it's needed for the `find_callsites` function here, https://github.com/JuliaDebug/Cthulhu.jl/blob/c37e281901642989effad82d9e2f0b0d803fe487/src/codeview.jl#L325

src/codeview.jl
```diff
@@ -144,7 +146,28 @@ function cthulhu_typed(io::IO, debuginfo::Symbol,
             if src.slottypes === nothing
                 @warn "Inference terminated in an incomplete state due to argument-type changes during recursion"
             end
-            printstyled(lambda_io, tsn; type_annotations, iswarn, hide_type_stable, idxend)
+            if TypedSyntax.isvscode()
```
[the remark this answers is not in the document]
Zentrik: I haven't been able to simplify this part of the code unfortunately by switching to an `IOContext`. Hopefully, it's a bit less messy now though.

pfitzseb:
> As Cthulhu already prints a hyperlink to the current function and because it seems like it would be annoying if the editor jumped everytime Cthulhu was used, for now I haven't implemented the second suggestion though I would be happy to if more people wanted it.

I feel like the redundancy of having the whole function printed in the REPL is more annoying than having the editor jump to the function you've descended into -- that workflow matches the normal debugging experience pretty closely. We could maybe add a `[F]ollow in editor` toggle?

At some point I'd also love to do this "properly" and use the debugger UI for this (additionally printing IR/native code in a separate pane to the side), but that's a bigger project.

Zentrik:
> he redundancy of having the whole function printed in the REPL is more annoying than having the editor jump to the function you've descended into -- that workflow matches the normal debugging experience pretty closely. We could maybe add a `[F]ollow in editor` toggle?

Is there a way to have the editor jump to a file without switching focus away from the terminal, I presume this is the desired behavior as otherwise it makes Cthulhu a bit annoying to use. I see that https://code.visualstudio.com/api/references/vscode-api#window.showTextDocument allows the terminal to preserve focus, though this doesn't seem to be exposed in the julia extension.

TypedSyntax/src/show.jl
```diff
@@ -55,6 +55,14 @@ Base.printstyled(rootnode::MaybeTypedSyntaxNode; kwargs...) = printstyled(stdout
 
 ndigits_linenumbers(node::AbstractSyntaxNode, idxend = last_byte(node)) = ndigits(node.source.first_line + nlines(node.source, idxend) - 1)
 
+function _print(io::IO, x, node, position)
+    print(io, x)
+
+    if !isempty(x) && position > 0 # position > 0 hacky fix not sure what the actual bug is
```
timholy, line 61: Not certain what this is about

timholy (commented):
This is looking very good to me. Once others who know more about the vscode side are happy with it, I'm fine with merging.

One question: how VSCode-specific is this, vs being LanguageServer-focused? Your PR here justifiably got a mention on `#appreciation` on the Julia slack, and one of the only concerns was about the tightness of the coupling. I'm not sure of the answer myself.

pfitzseb:
> this doesn't seem to be exposed in the julia extension.

Yeah, true. The focus switching would be super annoying for interactive use.

Zentrik:
> One question: how VSCode-specific is this, vs being LanguageServer-focused? Your PR here justifiably got a mention on `#appreciation` on the Julia slack, and one of the only concerns was about the tightness of the coupling. I'm not sure of the answer myself.

This doesn't use the LanguageServer at all, I don't think its possible to use LanguageServer here as that would require LanguageServer to load and compile user code. The vscode specific functions are in https://github.com/Zentrik/Cthulhu.jl/blob/master/TypedSyntax/src/vscode.jl and they are `Base.show` and those with vscode in their name. I think the main work to add this to other IDEs would be doing something like https://github.com/julia-vscode/julia-vscode/pull/3328 and the related code that was already in the extension.

timholy (approved):

Zentrik:
> I feel like the redundancy of having the whole function printed in the REPL is more annoying than having the editor jump to the function you've descended into -- that workflow matches the normal debugging experience pretty closely. We could maybe add a `[F]ollow in editor` toggle?

I've got this working, I'll make a pr once this is merged as it's largely separate of this pr and not VSCode specific.

TypedSyntax/src/vscode.jl
```diff
@@ -0,0 +1,91 @@
+diagnostics_available_vscode() = isdefined(Main, :VSCodeServer) && Main.VSCodeServer isa Module && isdefined(Main.VSCodeServer, :DIAGNOSTICS_ENABLED) && Main.VSCodeServer.DIAGNOSTICS_ENABLED[]
+inlay_hints_available_vscode() = isdefined(Main, :VSCodeServer) && Main.VSCodeServer isa Module && isdefined(Main.VSCodeServer, :INLAY_HINTS_ENABLED) && Main.VSCodeServer.INLAY_HINTS_ENABLED[]
```
vchuravy, line 2: @pfitzseb VSCodeServer is not registered right?

Otherwise we could use package extensions here.

pfitzseb: It is not registered, yes. Do package extensions actually require that though? VSCodeServer is a package and is loaded as such, so the appropriate hooks should fire.

vtjnash:
I am getting an error now after Pkg.update from this PR. It seems like the compat bounds are incorrect for these packages now, since I have Cthulhu dev'd but the released version of `TypedSyntax`, and those are not compatible anymore.

```
(@v1.8) pkg> st -m Cthulhu TypedSyntax
Status `~/.julia/environments/v1.8/Manifest.toml`
  [f68482b8] Cthulhu v2.9.2 `https://github.com/JuliaDebug/Cthulhu.jl.git#master`
  [d265eb64] TypedSyntax v1.2.1
```

```
julia> Base.locate_package(Base.identify_package(Base.identify_package(Main, "Cthulhu"), "TypedSyntax"))
"/home/vtjnash/.julia/packages/TypedSyntax/vbZHi/src/TypedSyntax.jl"

julia> Base.locate_package(Base.identify_package(Main, "Cthulhu"))
"/home/vtjnash/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl"
```

```
julia> using Cthulhu                                                                                                                                                                                      
Precompiling Cthulhu                                                                                                                                                                                      
        Info Given Cthulhu was explicitly requested, output will be shown live                                                                                                                            
ERROR: LoadError: UndefVarError: `clear_all_vscode` not defined                                                                                                                                           
Stacktrace:                                                                                                                                                                                               
  [1] getproperty(x::Module, f::Symbol)                                                                                                                                                                   
    @ Base ./Base.jl:31                                                                                                                                                                                   
  [2] __descend_with_error_handling(args::Any; terminal::Any, kwargs...)                                                                                                                                  
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:213                                                                                                                                          
  [3] _descend_with_error_handling(f::Any, argtypes::Any; kwargs::@Kwargs{iswarn::Bool, terminal::REPL.Terminals.TTYTerminal})                                                                            
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:200                                                                                                                                          
  [4] descend_code_typed(::Any, ::Vararg{Any}; kwargs::@Kwargs{terminal::REPL.Terminals.TTYTerminal})                                                                                                     
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:158                                                                                                                                          
  [5] macro expansion                                                                                                                                                                                     
    @ ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:831 [inlined]                                                                                                                                        
  [6] macro expansion                                                                                                                                                                                     
    @ ~/.julia/packages/PrecompileTools/kmH5L/src/workloads.jl:78 [inlined]                                                                                                                               
  [7] macro expansion                                                                                                                                                                                     
    @ ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:830 [inlined]                                   
  [8] macro expansion                                                                                                                                                                                     
    @ ~/.julia/packages/PrecompileTools/kmH5L/src/workloads.jl:140 [inlined]                                                                                                                              
  [9] top-level scope                                                                                                                                                                                     
    @ ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:826                                                                                                                                                  
 [10] include                                                                                                                                                                                             
    @ Base ./Base.jl:494 [inlined]                                                                                                                                                                        
 [11] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, so
urce::Nothing)                                                                                                                                                                                            
    @ Base ./loading.jl:2216                                                                         
 [12] top-level scope                                                                                                                                                                                     
    @ stdin:3                                                                                                                                                                                             
in expression starting at /home/vtjnash/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:1                                                                                                                    
in expression starting at stdin:3                                                                                                                                                                         
                                                                                                                                                                                                          
caused by: UndefVarError: `diagnostics_available_vscode` not defined                                                                                                                                      
Stacktrace:                                                                                                                                                                                               
  [1] getproperty(x::Module, f::Symbol)                                                                                                                                                                   
    @ Base ./Base.jl:31                                                                              
  [2] cthulhu_typed(io::IOContext{IOBuffer}, debuginfo::Symbol, src::Core.CodeInfo, rt::Any, effects::Core.Compiler.Effects, mi::Core.MethodInstance; iswarn::Bool, hide_type_stable::Bool, optimize::Bool
, pc2remarks::Nothing, pc2effects::Nothing, inline_cost::Bool, type_annotations::Bool, annotate_source::Bool, inlay_types_vscode::Bool, diagnostics_vscode::Bool, interp::Cthulhu.CthulhuInterpreter)
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/codeview.jl:152                                                                                                                                         
  [3] cthulhu_typed(::IOContext{IOBuffer}, ::Cthulhu.DInfo.DebugInfo, ::Core.CodeInfo, ::Vararg{Any}; kwargs::@Kwargs{iswarn::Bool, optimize::Bool, hide_type_stable::Bool, pc2remarks::Nothing, pc2effect
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/codeview.jl:121                                                                                                                                         
  [4] (::Cthulhu.var"#92#99"{Cthulhu.CthulhuInterpreter, Cthulhu.DInfo.DebugInfo, Core.CodeInfo, Core.CodeInfo, DataType, Core.MethodInstance, Bool, Bool, Nothing, Nothing, Bool, Bool})(lambda_io::IOCon
text{IOBuffer})                                                                                                                                                                                           
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:481                                                                                                                                          
  [5] stringify(f::Any, context::IOContext{Base.DevNull})                                                                                                                                                 
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/ui.jl:90                                                                                                                                                
  [6] _descend(term::REPL.Terminals.TTYTerminal, interp::Cthulhu.CthulhuInterpreter, curs::Cthulhu.CthulhuCursor; override::Nothing, debuginfo::Symbol, optimize::Bool, interruptexc::Bool, iswarn::Bool, 
hide_type_stable::Bool, verbose::Nothing, remarks::Bool, with_effects::Bool, inline_cost::Bool, type_annotations::Bool, annotate_source::Bool, inlay_types_vscode::Bool, diagnostics_vscode::Bool)        
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:480                                                                                                                                          
  [7] _descend(term::REPL.Terminals.TTYTerminal, interp::Cthulhu.CthulhuInterpreter, mi::Core.MethodInstance; kwargs::@Kwargs{iswarn::Bool})                                                              
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:739                                                                                                                                          
  [8] _descend(term::REPL.Terminals.TTYTerminal, args::Any; interp::Core.Compiler.NativeInterpreter, kwargs::@Kwargs{iswarn::Bool})                                                                       
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:755                                                                                                                                          
  [9] __descend_with_error_handling(args::Any; terminal::Any, kwargs...)                                                                                                                                  
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:211                                                                                                                                          
 [10] _descend_with_error_handling(f::Any, argtypes::Any; kwargs::@Kwargs{iswarn::Bool, terminal::REPL.Terminals.TTYTerminal})                                                                            
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:200                                                                                                                                          
 [11] descend_code_typed(::Any, ::Vararg{Any}; kwargs::@Kwargs{terminal::REPL.Terminals.TTYTerminal})                                                                                                     
    @ Cthulhu ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:158                                                                                                                                          
 [12] macro expansion                                                                                                                                                                                     
    @ ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:831 [inlined]                                                                                                                                        
 [13] macro expansion                                                                                                                                                                                     
    @ ~/.julia/packages/PrecompileTools/kmH5L/src/workloads.jl:78 [inlined]                                                                                                                               
 [14] macro expansion                                                                                                                                                                                     
    @ ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:830 [inlined]                                                                                                                                        
 [15] macro expansion                                                                                                                                                                                     
    @ ~/.julia/packages/PrecompileTools/kmH5L/src/workloads.jl:140 [inlined]                                                                                                                              
 [16] top-level scope                                                                                                                                                                                     
    @ ~/.julia/packages/Cthulhu/stXnv/src/Cthulhu.jl:826                                                                                                                                                  
 [17] include                                                                                                                                                                                             
    @ Base ./Base.jl:494 [inlined]                                                                                                                                                                        
 [18] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, so
urce::Nothing)                                                                                                                                                                                            
    @ Base ./loading.jl:2216                                                                                                                                                                              
 [19] top-level scope                                                                                                                                                                                     
    @ stdin:3                                                                                        
```

test/runtests.jl
```diff
@@ -26,3 +26,60 @@ using Test, PerformanceTestTools
...
+
+    @testset "test_terminal.jl" begin
+        include("test_terminal.jl")
+    end
+
+    @testset "test_AbstractInterpreter.jl" begin
+        include("test_AbstractInterpreter.jl")
+    end
+end
+end
```
aviatesk, line 85: @Zentrik What's the whole purpose of duplicating tests here? Can't we `include("test_codeview_vscode.jl")` only?

Zentrik: There's some additional work that Cthulhu does when running under VSCode that I wanted to test in more cases than `"test_codeview_vscode.jl"`. The main thing I was worried about was descending into all callees [here](https://github.com/JuliaDebug/Cthulhu.jl/blob/33c237996f1c67764ce8d80df53e0593fb2ff8b9/src/codeview.jl#L161-L168). 
I don't think it's particularly important to run these additional tests given the code has been working for a while, so if you want you can nix it.

aviatesk: So, it sounds like the purpose of running these duplicated tests was to verify that descending into all callees doesn't lead to any timing issues, right? If so, we might consider removing them. Alternatively, if we decide to keep them, it could be a better design choice to extract the main portion of `runtests.jl` into a separate file and then `include` that file in both `runtests.jl` and `test_VSCode.jl`.

el-oso:
I see that this was merged into master, how can I test it myself?

Zentrik:
Call `Cthulhu.@descend ...` in the VSCode Julia REPL and it should just work.

el-oso:
This is very cool! thanks! 
Side question: Does this work in nvim? :-)

Zentrik:
No VSCode only unfortunately.

PR #477, opened 2023-07-16
Thread kwargs through code_llvm

gaurav-arya:
Fix use of keyword arguments when calling `code_llvm` on a bookmark without explicit IO


aviatesk (approved):
LGTM. Thanks!

PR #483, opened 2023-08-14
julia #50868 renamed the inst field of an Instruction to stmt for consistency

oscardssmith:
I'm not 100% sure this is optimal but seems to fix things.


vchuravy:
Thanks. Normally I would factor that out into a helper function.

PR #485, opened 2023-08-20
Add option to jump to source automatically

Zentrik:
This rebases and fixes #336 and for VSCode preserves focus on terminal.
This also hides the annotated source code print out in the REPL if we jump to the annotated source code.


timholy (approved):
The implementation looks fine.

Issue #492, opened 2023-08-30
crash in showing source code


vtjnash:
```
julia> using Cthulhu

julia> descend(Base._tuple_unique_fieldtypes, (Any,))
_tuple_unique_fieldtypes(t) @ Base tuple.jl:212
212 function _tuple_unique_fieldtypes(@nospecialize t::Any)::Core.SimpleVector
213     @_total_meta
214     types::Base.IdSet{Any} = IdSet::Type{Base.IdSet}()::Base.IdSet{Any}
215     t´::Any = unwrap_unionall(t::Any)::Any
216     # Given t = Tuple{Vararg{S}} where S<:Real, the various
217     # unwrapping/wrapping/va-handling here will return Real
218     if (t´::Any isa Union)::Bool
219         union!(types::Base.IdSet{Any}, _tuple_unique_fieldtypes(rewrap_unionall(t´::Any.a::Any, t::Any)::Any)::Core.SimpleVector)
220         union!(types::Base.IdSet{Any}, _tuple_unique_fieldtypes(rewrap_unionall(t´::Any.b::Any, t::Any)::Any)::Core.SimpleVector)
221     else
222         for ti::Any in (ERROR: BoundsError: attempt to access Core.SimpleVector at index [1]
Stacktrace:
  [1] getindex(v::Core.SimpleVector, i::Int64)
    @ Base ./essentials.jl:769
  [2] type_annotation_mode(node::JuliaSyntax.TreeNode{TypedSyntax.TypedSyntaxData}, T::Any; type_annotations::Bool, hide_type_stable::Bool)
    @ TypedSyntax ~/.julia/dev/Cthulhu/TypedSyntax/src/show.jl:104
  [3] type_annotation_mode
    @ TypedSyntax ~/.julia/dev/Cthulhu/TypedSyntax/src/show.jl:97 [inlined]
  [4] show_src_expr(io::IOContext{Base.TTY}, node::JuliaSyntax.TreeNode{TypedSyntax.TypedSyntaxData}, position::Int64, pre::String, pre2::String; type_annotations::Bool, iswarn::Bool, hide_type_stable::Bool, nd::Int64)
    @ TypedSyntax ~/.julia/dev/Cthulhu/TypedSyntax/src/show.jl:79
[...]

PR #493, opened 2023-09-01
Check Type has parameters field with an element

### LINKED ISSUE #491: Some Types don't have field parameters error when checking if annotation is of form "String::Type{String}"
```jl
using StrideArraysCore, StrideArrays

julia> @inline function smallarray4(::Type{T}, ::StaticInt{SpillSize}, size::Integer...) where {T, SpillSize}
           @assert isbitstype(T)
           N = prod(size)
           if N <= SpillSize
               data = Ref{NTuple{SpillSize, T}}()
               p = Ptr{T}(pointer_from_objref(data))
           else
               data = Vector{T}(undef, N)
               p = Ptr{T}(pointer(data))
           end
           PtrArray(p, size, Vector{T}(undef, 32))
       end
smallarray4 (generic function with 1 method)

julia> @descend smallarray4(Float32, static(16), 16)
smallarray4(::Type{T}, ::StaticInt{SpillSize}, size::Integer...) where {T, SpillSize} @ Main REPL[210]:1
┌ Warning: Some line information is missing, type-assignment may be incomplete
└ @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/codeview.jl:144
ERROR: type TypeofBottom has no field parameters
Stacktrace:
  [1] getproperty
    @ Base ./Base.jl:32 [inlined]
  [2] type_annotation_mode(node::JuliaSyntax.TreeNode{…}, T::Any; type_annotations::Bool, hide_type_stable::Bool)
    @ TypedSyntax ~/.julia/packages/TypedSyntax/ByLzZ/src/show.jl:104
  [3] type_annotation_mode
    @ TypedSyntax ~/.julia/packages/TypedSyntax/ByLzZ/src/show.jl:97 [inlined]
  [4] printstyled(io::IOContext{…}, rootnode::JuliaSyntax.TreeNode{…}; type_annotations::Bool, iswarn::Bool, hide_type_stable::Bool, with_linenumber::Bool, idxend::Int64)
    @ TypedSyntax ~/.julia/packages/TypedSyntax/ByLzZ/src/show.jl:45
  [5] cthulhu_typed(io::IOContext{…}, debuginfo::Symbol, src::Core.CodeInfo, rt::Any, effects::Core.Compiler.Effects, mi::Core.MethodInstance; iswarn::Bool, hide_type_stable::Bool, optimize::Bool, pc2remarks::Nothing, pc2effects::Nothing, inline_cost::Bool, type_annotations::Bool, annotate_source::Bool, inlay_types_vscode::Bool, diagnostics_vscode::Bool, interp::Cthulhu.CthulhuInterpreter)
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/codeview.jl:168
  [6] cthulhu_typed(::IOContext{…}, ::Cthulhu.DInfo.DebugInfo, ::Core.CodeInfo, ::Vararg{…}; kwargs::@Kwargs{…})
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/codeview.jl:121
  [7] (::Cthulhu.var"#92#99"{…})(lambda_io::IOContext{…})
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/Cthulhu.jl:481
  [8] stringify(f::Any, context::IOContext{Base.TTY})
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/ui.jl:90
  [9] 
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/Cthulhu.jl:480
 [10] _descend(term::REPL.Terminals.TTYTerminal, interp::Cthulhu.CthulhuInterpreter, mi::Core.MethodInstance; kwargs::@Kwargs{…})
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/Cthulhu.jl:739
 [11] _descend(term::REPL.Terminals.TTYTerminal, args::Any; interp::Core.Compiler.NativeInterpreter, kwargs::@Kwargs{…})
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/Cthulhu.jl:755
 [12] __descend_with_error_handling(args::Any; terminal::Any, kwargs...)
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/Cthulhu.jl:211
 [13] _descend_with_error_handling(f::Any, argtypes::Any; kwargs::@Kwargs{iswarn::Bool})
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/Cthulhu.jl:200
 [14] descend_code_typed(::Any, ::Vararg{Any}; kwargs::@Kwargs{})
    @ Cthulhu ~/.julia/packages/Cthulhu/DxawF/src/Cthulhu.jl:158
 [15] top-level scope
    @ REPL[216]:1
Some type information was truncated. Use `show(err)` to see complete types.
```
Probably related to the function throwing
```jl
julia> smallarray4(Float32, static(16), 16)
ERROR: MethodError: no method matching (PtrArray)(::Ptr{Float32}, ::Tuple{Int64}, ::Vector{Float32})

Closest candidates are:
  (PtrArray)(::Ptr{T}, ::S, ::Val{R}) where {T, N, S<:Tuple{Vararg{Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8, StaticInt}, N}}, R}
   @ StrideArraysCore ~/.julia/packages/StrideArraysCore/COJRJ/src/ptr_array.jl:307
  (PtrArray)(::Ptr{T}, ::Tuple{Vararg{Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8, StaticInt}, N}}, ::Tuple{Vararg{Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8, StaticInt}, N}}, ::Val{D}) where {T, N, D}
   @ StrideArraysCore ~/.julia/packages/StrideArraysCore/COJRJ/src/ptr_array.jl:556
  (PtrArray)(::Ptr{T}, ::S, ::X, ::Val{R}) where {T, N, S<:Tuple{Vararg{Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8, StaticInt}, N}}, X<:Tuple{Vararg{Any, N}}, R}
   @ StrideArraysCore ~/.julia/packages/StrideArraysCore/COJRJ/src/ptr_array.jl:325
```

Zentrik:
Fixes #491, #492, #494

### DIFF
--- a/TypedSyntax/src/show.jl
+++ b/TypedSyntax/src/show.jl
@@ -99,7 +99,7 @@ function type_annotation_mode(node, @nospecialize(T); type_annotations::Bool, hi
     type_annotate = is_show_annotation(T; type_annotations, hide_type_stable)
     pre = pre2 = post = ""
     if type_annotate
-        if T isa Type && T <: Type
+        if T isa Type && T <: Type && T !== Union{} && isassigned(T.parameters, 1)
             # Don't annotate `String::Type{String}`
             if replace(sourcetext(node), r"\s" => "") == replace(sprint(show, T.parameters[1]), r"\s" => "")
                 return false, pre, pre2, post
--- a/TypedSyntax/src/show.jl
+++ b/TypedSyntax/src/show.jl
@@ -94,17 +94,17 @@ function is_show_annotation(@nospecialize(T); type_annotations::Bool, hide_type_
     return isa(T, Type) && is_type_unstable(T)
 end
 
+extract_inner_type(x) = nothing
+extract_inner_type(::Type{Type{T}}) where T = T
 function type_annotation_mode(node, @nospecialize(T); type_annotations::Bool, hide_type_stable::Bool)
     kind(node) == K"return" && return false, "", "", ""
     type_annotate = is_show_annotation(T; type_annotations, hide_type_stable)
     pre = pre2 = post = ""
     if type_annotate
-        if T isa Type && T <: Type && T !== Union{} && isassigned(T.parameters, 1)
-            # Don't annotate `String::Type{String}`
-            if replace(sourcetext(node), r"\s" => "") == replace(sprint(show, T.parameters[1]), r"\s" => "")
-                return false, pre, pre2, post
-            end
-        end
+        inner_type = extract_inner_type(T)
+        if T !== nothing && replace(sourcetext(node), r"\s" => "") == replace(sprint(show, inner_type), r"\s" => "")
+            return false, pre, pre2, post
+        end 
         if kind(node) ∈ KSet":: where" || is_infix_op_call(node) || (is_prec_assignment(node) && kind(node) != K"=")
             pre, post = "(", ")"
         elseif is_prefix_op_call(node) # insert parens after prefix op and before type-annotating
--- a/TypedSyntax/test/runtests.jl
+++ b/TypedSyntax/test/runtests.jl
@@ -650,6 +650,16 @@ include("test_module.jl")
     @test_nowarn str = sprint(tsn; context=:color=>false) do io, obj
         printstyled(io, obj; hide_type_stable=false)
     end
+
+    # issue 491
+    tsn = TypedSyntaxNode(+, (Int, Int)) # need a node, not important what it is
+    @test_nowarn TypedSyntax.type_annotation_mode(tsn, Union{}; type_annotations=true, hide_type_stable=false)
+
+    # issue 492
+    tsn = TypedSyntaxNode(Base._tuple_unique_fieldtypes, (Any,))
+    @test_nowarn str = sprint(tsn; context=:color=>false) do io, obj
+        printstyled(io, obj; hide_type_stable=false)
+    end
 end
 
 if parse(Bool, get(ENV, "CI", "false"))
--- a/TypedSyntax/test/runtests.jl
+++ b/TypedSyntax/test/runtests.jl
@@ -660,6 +660,12 @@ include("test_module.jl")
     @test_nowarn str = sprint(tsn; context=:color=>false) do io, obj
         printstyled(io, obj; hide_type_stable=false)
     end
+
+    # issue 493
+    tsn = TypedSyntaxNode(TSN.f493, ())
+    @test_nowarn str = sprint(tsn; context=:color=>false) do io, obj
+        printstyled(io, obj; hide_type_stable=false)
+    end
 end
 
 if parse(Bool, get(ENV, "CI", "false"))
--- a/TypedSyntax/test/test_module.jl
+++ b/TypedSyntax/test/test_module.jl
@@ -237,4 +237,9 @@ f482b(x) = Dict{String,Any}(x)
 # Issue 487
 f487(x) = 1
 
+function f493()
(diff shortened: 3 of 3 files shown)


Zentrik:
Unfortunately it seems that `extract_inner_type(::Type{Type{T}})` specializes on each new type it sees, I don't know how problematic that is. Also not sure why we were taking the first element of `T.parameters` before but this implementation seems to work fine.

TypedSyntax/src/show.jl
```diff
@@ -94,17 +94,17 @@ function is_show_annotation(@nospecialize(T); type_annotations::Bool, hide_type_
...
 end
 
+extract_inner_type(x) = nothing
+extract_inner_type(::Type{Type{T}}) where T = T
 function type_annotation_mode(node, @nospecialize(T); type_annotations::Bool, hide_type_stable::Bool)
     kind(node) == K"return" && return false, "", "", ""
     type_annotate = is_show_annotation(T; type_annotations, hide_type_stable)
     pre = pre2 = post = ""
     if type_annotate
-        if T isa Type && T <: Type
```
timholy, line 102: Hmm, I guess this should have been `T <: Type{S} where S` or something?

timholy (commented):
I doubt it will be measurable (and so we can merge this), but presumably `extract_inner_type` will be specialized for every type. The old code tried to avoid that but it may not be readily fixable. So I'm fine with merging this, thanks for tackling it!

timholy:
> Also not sure why we were taking the first element of `T.parameters` before but this implementation seems to work fine.

```julia
julia> TT = Type{Float64}
Type{Float64}

julia> TT.parameters[1]
Float64
```

Basically we were extracting the `T` in the same way as your `extract_inner_type`, without forcing dispatch specialization. I think the problem with the old implementation is that `T <: Type` was too weak of a condition. Perhaps `T <: Type{S} where S` might fix it?

TypedSyntax/src/show.jl
Zentrik: `Union{} <: Type{S} where S` is true so I don't think that would have fixed it either.

Zentrik:
I thought `extract_inner_type(@nospecialize(x::Type{<:Type})) = x.parameters[1]` might work, but unfortunately it fails with `Core.TypeofBottom`. Given I don't know when a type actually has a non empty `parameters` `svec` might be easiest to just use this implementation.

simeonschaub:
I believe it should be enough to just change the check to `T isa DataType && T <: Type`

Zentrik:
`Core.TypeofBottom` satisfies that but has an empty parameters svec, maybe checking if it's non empty as well will work.

Zentrik:
Given we don't think the current implementation will be particularly slow I'll merge it for now, and we can revisit this if it is slow.

TypedSyntax/src/show.jl
vtjnash: It should be `T isa DataType` or `Base.isType(T)`

Issue #524, opened 2023-12-27
Precompilation fails on 32-bit Julia 1.9


ThatcherC:
I'm having an issue where I can't add Cthulhu to an environment in the Julia 1.9 32-bit version (`1.9~x86` channel on `juliaup`). It seems like it's the precompilation step that's failing. I've copied the results from trying to add and precompile Cthulhu down below.

For context, I'm hoping to use Cthulhu to resolve a type issue I encountered while trying to use GPUCompiler to compile a function down to WASM based on the approached taken by @Alexander-Barth's [FluidSimDemo-WebAssembly](https://github.com/Alexander-Barth/FluidSimDemo-WebAssembly#generating-wasm-binary-from-julia-code) -- that's the only reason I'm using the 32-bit x86 release. I think Cthulhu is exactly the tool I need to track down this other type issue so I hope there's a way to get it to work on the x86 release!

Here's the error I'm seeing:

```
$ julia +1.9~x86
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.4 (2023-11-14)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(@v1.9) pkg> add Cthulhu
   Resolving package versions...
    Updating `~/.julia/environments/v1.9/Project.toml`
  [f68482b8] + Cthulhu v2.10.0
    Updating `~/.julia/environments/v1.9/Manifest.toml`
  [da1fd8a2] + CodeTracking v1.3.5
  [f68482b8] + Cthulhu v2.10.0
  [1eca21be] + FoldingTrees v1.2.1
  [70703baa] + JuliaSyntax v0.4.8
  [d265eb64] + TypedSyntax v1.2.3
  [b8c1c048] + WidthLimitedIO v1.0.1
Precompiling project...
  ✗ Groebner
  ✗ Symbolics
  0 dependencies successfully precompiled in 53 seconds. 370 already precompiled. 1 skipped during auto due to previous errors.
  2 dependencies errored. To see a full report either run `import Pkg; Pkg.precompile()` or load the packages

(@v1.9) pkg> precompile Cthulhu
Precompiling Cthulhu
  ✗ Cthulhu
[...]

PR #529, opened 2024-01-07
Add x86 runner to TypedSyntax CI

Zentrik:
It seems like #524 probably has existed for a long time, having CI would probably be helpful then.


Zentrik:
I believe the nightly failures are due to https://github.com/JuliaLang/julia/issues/52796.

Zentrik:
Pretty sure the other failures are unrelated.

Issue #494, opened 2023-09-03
type Union has no field parameters


homocomputeris:
Running the [example](https://github.com/JuliaDebug/Cthulhu.jl?tab=readme-ov-file#usage-descend) from the readme
```julia
using Cthulhu

function foo()
    T = rand() > 0.5 ? Int64 : Float64
    sum(rand(T, 100))
end

@descend foo()            # option 2: apply `@descend` to a working execution of the function
```
raises
```julia
foo() @ Main ~/Code/test/test.jl:3
ERROR: type Union has no field parameters
Stacktrace:
  [1] getproperty
    @ ./Base.jl:32 [inlined]
  [2] type_annotation_mode(node::JuliaSyntax.TreeNode{TypedSyntax.TypedSyntaxData}, T::Any; type_annotations::Bool, hide_type_stable::Bool)
    @ TypedSyntax ~/.julia/packages/TypedSyntax/ByLzZ/src/show.jl:104
  [3] type_annotation_mode
    @ ~/.julia/packages/TypedSyntax/ByLzZ/src/show.jl:97 [inlined]
  [4] show_src_expr(io::IOContext{IOBuffer}, node::JuliaSyntax.TreeNode{TypedSyntax.TypedSyntaxData}, position::Int64, pre::String, pre2::String; type_annotations::Bool, iswarn::Bool, hide_type_stable::Bool, nd::Int64)
    @ TypedSyntax ~/.julia/packages/TypedSyntax/ByLzZ/src/show.jl:79
  [5] show_src_expr(io::IOContext{IOBuffer}, node::JuliaSyntax.TreeNode{TypedSyntax.TypedSyntaxData}, position::Int64, pre::String, pre2::String; type_annotations::Bool, iswarn::Bool, hide_type_stable::Bool, nd::Int64)
    @ TypedSyntax ~/.julia/packages/TypedSyntax/ByLzZ/src/show.jl:80
  [6] show_src_expr
    @ ~/.julia/packages/TypedSyntax/ByLzZ/src/show.jl:66 [inlined]
  [7] printstyled(io::IOContext{IOBuffer}, rootnode::JuliaSyntax.TreeNode{TypedSyntax.TypedSyntaxData}; type_annotations::Bool, iswarn::Bool, hide_type_stable::Bool, with_linenumber::Bool, idxend::Int64)
    @ TypedSyntax ~/.julia/packages/TypedSyntax/ByLzZ/src/show.jl:50
[...]

PR #537, opened 2024-01-17
Create overloadable function for string representations of types

### LINKED ISSUE #535: Allow user to overload printing behavior
(x-ref https://discourse.julialang.org/t/proposed-alias-for-union-types/108205/142?u=milescranmer)

Julia’s type system is the best I’ve encountered in any language, but it can admittedly get a bit verbose at times. Especially as some types stretch over the VSCode length limit in the Cthulhu labels and end up missing some information.

Therefore I was wondering if Cthulhu could permit the user to overload printing behavior for types in a Cthulhu printout, without necessarily overloading the printing behavior for all of Julia, like `Float32 → F32`, `Tuple → Tu`, etc, just to make the printouts less verbose.

For example:

```julia
import Cthulhu: type_string

## default defined in Cthulhu:
# type_string(t) = string(t)

# user aliases:
type_string(::Type{Float32}) = "F32"
type_string(::Type{UInt8}) = "U8"
type_string(::Type{Union{A,B}}) where {A,B} = "(" * type_string(A) * "|" * type_string(B) * ")"
# etc.
```

which you could have in your startup.jl.

This would be backwards compatible and would simply allow the user to customize printouts as they wish.

Cheers,
Miles

MilesCranmer:
Fixes #535

### DIFF
--- a/TypedSyntax/src/show.jl
+++ b/TypedSyntax/src/show.jl
@@ -113,12 +113,14 @@ function type_annotation_mode(node, @nospecialize(T); type_annotations::Bool, hi
     return type_annotate, pre, pre2, post
 end
 
+type_string(T) = string(T)
+
 function show_annotation(io, @nospecialize(T), post, node, position; iswarn::Bool)
     diagnostics = get(io, :diagnostics, nothing)
     inlay_hints = get(io, :inlay_hints, nothing)
 
     print(io, post)
-    T_str = string(T)
+    T_str = type_string(T)
     if iswarn && is_type_unstable(T)
         color = is_small_union_or_tunion(T) ? :yellow : :red
         printstyled(io, "::", T_str; color)


TypedSyntax/src/show.jl
```diff
@@ -113,12 +113,21 @@ function type_annotation_mode(node, @nospecialize(T); type_annotations::Bool, hi
...
+"""
+type_string(T) = string(T)
+
 function show_annotation(io, @nospecialize(T), post, node, position; iswarn::Bool)
     diagnostics = get(io, :diagnostics, nothing)
     inlay_hints = get(io, :inlay_hints, nothing)
 
     print(io, post)
-    T_str = string(T)
+    T_str = type_string(T)
```
timholy, line 130: Do we need to make this recursive? I think it would be weird if we printed like this:
```julia
Float32: "F32"
Vector{Float32}: "Vector{Float32}"
```

timholy (commented):
Add a few tests?

TypedSyntax/src/show.jl
MilesCranmer: I think it might be hard to do that generally – I just assumed the user could set that up for all the types they'd be interested in. Otherwise it would get really complicated quickly. But maybe there's an easy way to apply it recursively?

MilesCranmer: Maybe we could set it up for the basic ones though, like `Vector`, `Union`, `Core.Const`, etc.?

timholy: Here's a basis for doing it systematically:

```julia
julia> T = Dict{String,Vector{Float32}}
Dict{String, Vector{Float32}}

julia> typeof(T)
DataType

julia> T.parameters
svec(String, Vector{Float32})

julia> T.parameters[1]
String

julia> T.parameters[2].parameters
svec(Float32, 1)
```

timholy: Alternative thought: generate the string and `replace`?

MilesCranmer: I see, let me try, thanks.

Re: strings, I've tried this in a different context but it ended up being a lot of trouble because regexp isn't Turing complete. For example there's no way to do nested parentheses matching with regexp. After that experience I've mostly stuck to converting to a string as late as possible.

MilesCranmer:
Okay I think I'm hitting some recursive types now. It's not fully general but the user can always overload behavior when needed.

For the tests I think I'm doing something wrong in that I don't understand `Cthulhu.cthulhu_typed`'s output and whether it actually triggers `type_string` or not. Mind taking a look?

TypedSyntax/src/show.jl
```diff
@@ -113,12 +113,40 @@ function type_annotation_mode(node, @nospecialize(T); type_annotations::Bool, hi
...
+`type_string(::Type{Float32}) = "F32"` would result in `F32` being used to annotate
+any instance of a `Float32`.
+"""
+function type_string(T)
+    if T isa DataType
+        if isempty(T.parameters)
+            return string(T)
+        else
+            wrapper = Base.typename(T).wrapper
+            return string(wrapper, '{', join(map(type_string, T.parameters), ','), '}')
```
Zentrik, line 129: I'm a bit wary of this code mapping over `T.parameters` as there were quite a few issues doing a similar thing elsewhere. 
In that code we had to check for whether each element in `T.parameters` was assigned, I don't know if that's a concern here or not.
See https://github.com/JuliaDebug/Cthulhu.jl/pull/508/files for that code and some issues with previous versions of that code are #491, #492, #494.

PR #520, opened 2023-11-26
Support tagged CodeInstance

vchuravy:
Support JuliaLang/julia#52233

PR #540, opened 2024-02-10
Support tagged CodeInstances (Valentin's version)

vchuravy:
Simpler version of #520 

We don't use the InternalCodeCache yet
since that has troubles with our augmented
inference returns.


vchuravy:
Tests pass locally.

PR #543, opened 2024-02-12
minor follow up on the tagged code instance change

aviatesk:
Requires JuliaLang/julia#53300.

### DIFF
--- a/src/interpreter.jl
+++ b/src/interpreter.jl
@@ -24,7 +24,7 @@ end
 
 const InferenceKey = Union{MethodInstance,InferenceResult}
 const InferenceDict{T} = IdDict{InferenceKey, T}
-const OptimizationDict = @static VERSION ≥ v"1.11.0-DEV.1552" ? Nothing : IdDict{MethodInstance, CodeInstance}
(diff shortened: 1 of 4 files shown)


aviatesk:
There appears to be remaining issues related to world ages of `CthulhuInterpreter`. In the past, code instances were simply inserted into a dictionary maintained by `CthulhuInterpreter` and the cache existence was simply equivalent to the key existence. However, as they are now maintained as what could be considered "proper" `CodeInstance`s, we're encountering complex problems in test cases that include `do` blocks."

Keno:
Do we actually want this, semantically? The fact that Cthulhu's cache is not poisoned by prior results is one of the best features about it. Also, I don't really see any benefit here. You're not likely to want to cache Cthulhu results in precompile files, etc.

vchuravy:
>Do we actually want this, semantically? 

Yeah I was thinking about this as well and that's why in #540 I kept using the old design. Each instance of the abstract interpreter currently get's a fresh cache.

aviatesk:
Yeah, I agree that the current design with the external code cache is a good fit for Cthulhu.

PR #545, opened 2024-02-13
drop v1.9 support

aviatesk:


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Cthulhu"
 uuid = "f68482b8-f384-11e8-15f7-abe071a5a75f"
 authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
-version = "2.10.2"
```
vchuravy, line 4: Doesn't this need to be `2.11`?

vchuravy (approved):

Project.toml
aviatesk: I've had difficulty deciding on how to handle semver when bumping Julia's own version compatibility. If a minor version bump appears to be more appropriate, I'm just fine with proceeding in that direction.

Issue #523, opened 2023-12-24
Internal `AssertionError: label > 0` when descending a method call


MilesCranmer:
I'm seeing this weird internal assertion error when inspecting one of my method calls. Here is the current MWE. I'm setting up a search problem in SymbolicRegression and trying to descend the main search method:

```julia
julia> using Cthulhu

julia> using SymbolicRegression

julia> X = rand(2, 1000) .* 30;

julia> y = @. sin(X[1, :]) + 1.5;

julia> options = Options(; binary_operators=[+, -, *, /, mod], unary_operators=[]);

julia> @descend equation_search(X, y; options)
```

This triggers the following error:

```julia
julia> @descend equation_search(X, y; options)
ERROR: AssertionError: label > 0
Stacktrace:
   [1] 
     @ Core.Compiler ./compiler/ssair/ir.jl:1321
   [2] cfg_simplify!(ir::Core.Compiler.IRCode)
     @ Core.Compiler ./compiler/ssair/passes.jl:2220
   [3] create_cthulhu_source
     @ Cthulhu ~/.julia/packages/Cthulhu/O1Xhq/src/interpreter.jl:158 [inlined]
   [4] transform_result_for_cache
     @ Cthulhu ~/.julia/packages/Cthulhu/O1Xhq/src/interpreter.jl:164 [inlined]
   [5] cache_result!(interp::Cthulhu.CthulhuInterpreter, result::Core.Compiler.InferenceResult)
     @ Core.Compiler ./compiler/typeinfer.jl:405
   [6] _typeinf(interp::Cthulhu.CthulhuInterpreter, frame::Core.Compiler.InferenceState)
     @ Core.Compiler ./compiler/typeinfer.jl:283
   [7] typeinf(interp::Cthulhu.CthulhuInterpreter, frame::Core.Compiler.InferenceState)
     @ Core.Compiler ./compiler/typeinfer.jl:219
   [8] 
     @ Core.Compiler ./compiler/typeinfer.jl:930
   [9] 
     @ Core.Compiler ./compiler/abstractinterpretation.jl:629
  [10] 
     @ Core.Compiler ./compiler/abstractinterpretation.jl:95
  [11] 
     @ Core.Compiler ./compiler/abstractinterpretation.jl:2083
  [12] 
     @ Core.Compiler ./compiler/abstractinterpretation.jl:2165
  [13] 
     @ Core.Compiler ./compiler/abstractinterpretation.jl:2158
[...]

PR #546, opened 2024-02-17
Allow user to overload cfg_simplify!

MilesCranmer:
Allows users to overload the method `Cthulhu.cfg_simplify!`

(**NOTE** `Cthulhu.cfg_simplify! != CC.cfg_simplify!` so this isn't messing with internals of the compiler).

This is so that power users can just turn it off in their `startup.jl` and avoid this super annoying bug: https://github.com/JuliaDebug/Cthulhu.jl/issues/523 which seems like we are just waiting on Julia 1.11 for a fix.

Here's how to use it. In your `startup.jl` file:

```julia
using Cthulhu  # Regular import

# Rewrite buggy code:
import Cthulhu: cfg_simplify!
cfg_simplify!(ir) = ir
```

### DIFF
--- a/src/interpreter.jl
+++ b/src/interpreter.jl
@@ -162,10 +162,13 @@ function CC.finish(state::InferenceState, interp::CthulhuInterpreter)
     return res
 end
 
+# Overload me to disable
+cfg_simplify!(ir) = CC.cfg_simplify!(ir)
+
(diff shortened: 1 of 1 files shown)


aviatesk:
I think the right fix here would be to disable `cfg_simplify!` on v1.10 at all, or replace it with the proper round-trip. See #548.

MilesCranmer:
SGTM, thanks!

PR #551, opened 2024-03-03
check CI on master

aviatesk:


aviatesk:
CI seems to be broken wildly. Looks like all failures are VSCode-related. I'm thinking of temporarily disabling those tests to ensure other development work can continue safely. Meanwhile, I'd like to request @Zentrik to address the issue when he has the opportunity.

PR #547, opened 2024-02-19
Adjust to JuliaLang/julia#53219

Keno:
These are the changes I was using locally when developing that branch to have basic Cthulhu functionality. It's likely additional changes are required to have Cthulhu fully functional.

- closes #552

PR #553, opened 2024-03-03
TypedSyntax.jl: drop old version support

aviatesk:
Given that Cthulhu is the only user of TypedSyntax currently, and considering the newer versions of Cthulhu no longer supports older Julia versions like 1.6, I believe there isn't much merit in maintaining support for older Julia versions within TypedSyntax. We anticipate the need for fixes to TypedSyntax's reflection capabilities, like #547, however, continuing to provide support for these features across older versions proves to be a significant hassle. Therefore, I'm inclined to match TypedSyntax.jl's supported Julia version with that of Cthulhu.

Additionally, we've found that if Cthulhu's tests fail, TypedSyntax.jl's tests don't get executed, leading to unnoticed regressions in the latter. So I've opted to separate and individually run the tests for both Cthulhu and TypedSyntax.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -43,17 +43,11 @@ jobs:
             ${{ runner.os }}-test-${{ env.cache-name }}-
             ${{ runner.os }}-test-
             ${{ runner.os }}-
-      - name: dev TypedSyntax
-        shell: julia --color=yes --project=. {0}    # this is necessary for the next command to work on Windows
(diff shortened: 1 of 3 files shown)


aviatesk:
I haven't received any reviews yet, but I'm considering merging this in order for #547 and other PRs to make progress. If anyone wants to keep support for versions 1.6~1.9, please inform me. I'm willing to take responsibility for any necessary reverts.

Note that the failures of TypedSyntaxCI on 1.11-nightly and nightly are not new regression, but existing ones (although we may have not perceived yet).

PR #526, opened 2024-01-03
changes to adapt to compressed line table format

vtjnash:
This contains some updates needed to keep working with the optimized debuginfo format (https://github.com/JuliaLang/julia/pull/52415).

Also needs https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/606

PR #558, opened 2024-03-19
adjustments to JuliaLang/julia#52415

aviatesk:
/cc @vtjnash

### DIFF
--- a/src/codeview.jl
+++ b/src/codeview.jl
@@ -68,9 +68,13 @@ end
 using Base.IRShow: IRShow, _stmt, _type, should_print_ssa_type, IRShowConfig, show_ir
 
 const __debuginfo = merge(IRShow.__debuginfo, Dict(
-    :compact => src -> src isa CodeInfo ? __debuginfo[:source](src)
-                                        : IRShow.inline_linfo_printer(src)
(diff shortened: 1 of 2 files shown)


aviatesk:
This PR leverages the new type printing feature from #52415:
image

vtjnash:
Seems to be a less complete version of https://github.com/JuliaDebug/Cthulhu.jl/pull/526? That was currently blocked on getting my PRs for JuliaInterpreter merged so it works on Julia v1.11+ again

I am excited to see the new type printing feature getting used, but I am skeptical that other folks will want to have doubled (or more) the number of contextual lines that get printed by default. It might need to be opt-in in some way?

aviatesk:
Oops, I didn't notice you'd already put in a PR. My bad. Anyway, why should we prioritize merging JuliaInterpreter? Since Cthulhu doesn't rely on JuliaInterpreter, we could update it on its own, right? It sounds like updating JuliaInterpreter might take a bit while.

aviatesk:
Regarding the extra debug information, I believe it should be okay because `:source`-mode is already an opt-in feature.

vtjnash:
I thought this relied on it. But maybe that was Revise instead that relies on it?

PR #562, opened 2024-04-08
Make it merely a warning for the precompile workload to fail

oxinabox:
Cthulhu breaks easily on nightly, but not everyone uses all Cthulhu features.
So making it only a warning while we fix things may be useful.
In particular, right now Cthulu is broken on 1.11-alpha2, due to #343  though it is fixed at the head of the release-1.11 branch.

This in turn breaks Diffractor which has some Cthulhu integrations (reasonable) and also reuses some Cthulhu internals (not so reasonable).
But which actually works fine, its just #343 

cf https://github.com/JuliaLang/PrecompileTools.jl/pull/21

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -876,13 +876,17 @@ ascend
 
 using PrecompileTools
 @setup_workload begin
-    input = Base.link_pipe!(Pipe(), reader_supports_async=true, writer_supports_async=true)
-    term = REPL.Terminals.TTYTerminal("dumb", input.out, devnull, devnull)
-    write(input.in, 'q')
-    @compile_workload begin
-        descend(gcd, (Int, Int); terminal=term)
-        # declare we are done with streams
-        close(input.in)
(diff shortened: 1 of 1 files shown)


aviatesk:
I'm in favor of this change, but still have a question: #343 is an issue that was fixed on the Cthulhu-side a while back. Why is it still leading to errors in the `release-1.11` Julia?

oxinabox:
I assume because a new alpha hasn't been cut since it was fixed?
I haven't actually tested the head of release-11, just alpha2 which is in juliaup

aviatesk:
Did you get the same error as https://github.com/JuliaDebug/Cthulhu.jl/issues/342?
I built `release-v1.11` Julia manually but I couldn't reproduce it on macOS on the latest Cthulhu master.

oxinabox:
oh, now that i look closer no i didn't.
I got a similar but slightly different error.

I will open an issue

PR #568, opened 2024-04-25
Use depth-limited type printing

### LINKED ISSUE #566: Use depth-limited type printing, from Julia 1.10
This is my analog request of https://github.com/aviatesk/JET.jl/issues/616

charleskawczynski:
Closes #566.

Right now, this isn't working for me locally, but @Zentrik suggested this is where we can put `type_depth_limit`.


Zentrik:
TypedSyntax is a separate package that can work without Cthulhu, so the `print.jl` file needs to be in TypedSyntax and can't rely on a global in Cthulhu.
It looks like print.jl is only a line or 2 so seems simplest to just put the code where `T_str = string(T)` is currently, otherwise you should probably add `@nospecialize` to the arguments given the function `T_str = string(T)` is contained in does.

charleskawczynski:
> TypedSyntax is a separate package that can work without Cthulhu, so the `print.jl` file needs to be in TypedSyntax and can't rely on a global in Cthulhu. It looks like print.jl is only a line or 2 so seems simplest to just put the code where `T_str = string(T)` is currently, otherwise you should probably add `@nospecialize` to the arguments of the new functions given the function `T_str = string(T)` is contained in does.

I still never managed to exercise `show_annotation`, so it's difficult for me to debug this, but moving it there seems like the right way forward, and we can then figure out how to pass `CONFIG.type_depth_limit` to it through `Base.printstyled`

charleskawczynski:
I think that this is now working. From the test added in the test suite:

Screen Shot 2024-04-25 at 12 53 27 PM

charleskawczynski:
On main, it looks like this:

Screen Shot 2024-04-25 at 12 55 39 PM

charleskawczynski:
I'm not sure what the best way is forward for testing, @vchuravy?

TypedSyntax/src/show.jl
```diff
@@ -113,12 +115,25 @@ function type_annotation_mode(node, @nospecialize(T); type_annotations::Bool, hi
...
+    type_depth_limit(io, string(T); maxtypedepth)
+end
+
+function show_annotation(io, @nospecialize(T), post, node, position; iswarn::Bool, maxtypedepth)
     diagnostics = get(io, :diagnostics, nothing)
     inlay_hints = get(io, :inlay_hints, nothing)
 
     print(io, post)
-    T_str = string(T)
+    T_str = type_depth_limit(T; maxtypedepth)
```
Zentrik, line 136: Why do we need 3 new functions, this works right? 
```
T_str = string(T)
if maxdepthlimit !== nothing
	sz = get(io, :displaysize, displaysize(io))::Tuple{Int, Int}
	T_str = Base.type_depth_limit(T_str, max(sz[2], 120); maxtypedepth)
end
```

Anyways, as far as I can tell only `max(sz[2], 120)` or `maxtypedepth` is used. It seems better to control the string width directly so perhaps we should control that instead.

src/codeview.jl
```diff
@@ -119,6 +119,7 @@ function cthulhu_typed(io::IO, debuginfo::Symbol,
     lineprinter = __debuginfo[debuginfo]
     rettype = ignorelimited(rt)
     lambda_io = IOContext(io, :limit=>true)
```
Zentrik, line 121: Can we  not add the max_type_depth to this as discussed at https://github.com/JuliaDebug/Cthulhu.jl/issues/566#issuecomment-2077648580

Zentrik (changes requested):

Zentrik:
Thanks for this, I've left a couple comments as these seems unnecessarily complicated right now.

TypedSyntax/src/show.jl
charleskawczynski: Done

src/codeview.jl
charleskawczynski: How would that be done?

Zentrik: Oh sorry, I didn't realise you were passing maxtypedepth to mutliple print functions.
Anways, I would assume you could do `IOContext(io, :maxtypedepth=>CONFIG.type_depth_limit)` at the callsites where you are passing maxtypedepth or just redefine `io` to be that.

charleskawczynski: Does `IOContext(io, :maxtypedepth=>CONFIG.type_depth_limit)` work? I don't see `maxtypedepth` in the docs:

```julia
help?> IOContext
search: IOContext

  IOContext

  IOContext provides a mechanism for passing output configuration settings among show methods.

  In short, it is an immutable dictionary that is a subclass of IO. It supports standard dictionary operations such as getindex, and can also be used as an I/O stream.

  ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  IOContext(io::IO, context::IOContext)

  Create an IOContext that wraps an alternate IO but inherits the properties of context.

  ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  IOContext(io::IO, KV::Pair...)

  Create an IOContext that wraps a given stream, adding the specified key=>value pairs to the properties of that stream (note that io can itself be an IOContext).

    •  use (key => value) in io to see if this particular combination is in the properties set

    •  use get(io, key, default) to retrieve the most recent value for a particular key

  The following properties are in common use:

    •  :compact: Boolean specifying that values should be printed more compactly, e.g. that numbers should be printed with fewer digits. This is set when printing array
       elements. :compact output should not contain line breaks.

    •  :limit: Boolean specifying that containers should be truncated, e.g. showing … in place of most elements.

    •  :displaysize: A Tuple{Int,Int} giving the size in rows and columns to use for text output. This can be used to override the display size for called functions, but to
       get the size of the screen use the displaysize function.

    •  :typeinfo: a Type characterizing the information already printed concerning the type of the object about to be displayed. This is mainly useful when displaying a
       collection of objects of the same type, so that redundant type information can be avoided (e.g. [Float16(0)] can be shown as "Float16[0.0]" instead of
       "Float16[Float16(0.0)]" : while displaying the elements of the array, the :typeinfo property will be set to Float16).

    •  :color: Boolean specifying whether ANSI color/escape codes are supported/expected. By default, this is determined by whether io is a compatible terminal and by any
       --color command-line flag when julia was launched.

  Examples
  ≡≡≡≡≡≡≡≡
```

Zentrik: You can pass arbitrary keys, e.g. https://github.com/JuliaDebug/Cthulhu.jl/blob/17c53a157f5c21793115dc53ff83c87147dbc47b/src/codeview.jl#L143-146

TypedSyntax/src/show.jl
```diff
@@ -113,12 +115,16 @@ function type_annotation_mode(node, @nospecialize(T); type_annotations::Bool, hi
...
-function show_annotation(io, @nospecialize(T), post, node, position; iswarn::Bool)
+function show_annotation(io, @nospecialize(T), post, node, position; iswarn::Bool, maxtypedepth)
     diagnostics = get(io, :diagnostics, nothing)
     inlay_hints = get(io, :inlay_hints, nothing)
 
     print(io, post)
     T_str = string(T)
+    if maxtypedepth !== nothing
+        sz = get(io, :displaysize, displaysize(io))::Tuple{Int, Int}
+        T_str = Base.type_depth_limit(T_str, max(sz[2], 120); maxdepth=maxtypedepth)
```
Zentrik, line 126: As far as I can tell only max(sz[2], 120) or maxtypedepth is used not both. It seems better to control the string width directly so perhaps we should control that instead.

src/codeview.jl
charleskawczynski: Interesting, sure I'll go ahead and pass `:maxtypedepth=>CONFIG.type_depth_limit`

TypedSyntax/src/show.jl
charleskawczynski: I think this may still be useful when `maxdepthtype = nothing`?

src/codeview.jl
charleskawczynski: Let me know if I've not done that correctly

Zentrik: Thanks, sorry but I think I forgot to explain why I wanted this which has probably caused some confusion.
Right now, you are passing maxtypedepth as the argument to lots of functions so that it's available in `show_annotation`.
The benefit to putting `maxtypedepth` in the io is that, the io is present in `show_annotation` so if you do `get(io, :maxtypedepth, nothing)` is `show_annotation` we can avoid passing `maxtypedepth` as an argument everywhere.
E.g.
```
io = IOContext(io, :a=>4)
f(io)

function f(io)
	g(io)
end

function g(io)
	a = get(io, :a, nothing)
	println(a)
end
```
Note that `a` didn't need to be passed into `f` or `g`. For another example, look how `vscode_io` is defined a couple lines below and how it is used in `show_annotation` to pass the `inlay_hints` and `diagnostics` variables.

Let me know if anything is unclear.

TypedSyntax/src/show.jl
Zentrik: Yes you're right, sorry. But to try to explain myself more clearly, your issue was that some types that are printed are extremely long. It seems to me the second argument to `Base.type_depth_limit` controls the size of the printed type so is there a reason we don't set a limit for that instead of limiting the type depth?

Zentrik:
For the tests, https://github.com/JuliaDebug/Cthulhu.jl/blob/master/test/test_codeview.jl looks like a good place to put it and hopefully the other tests in that file will help you in writing one for this.
E.g. I think you should be able to adapt https://github.com/JuliaDebug/Cthulhu.jl/blob/17c53a157f5c21793115dc53ff83c87147dbc47b/test/test_codeview.jl#L49-L74 to test this.

TypedSyntax/src/show.jl
charleskawczynski: Yes, but for packages like JET/Cthulhu, it's sometimes helpful to set this depth explicitly. For example, in closures, the depth is too shallow and you cannot see types that you would like to.

src/codeview.jl
charleskawczynski: Sounds good, I tried this out and I think it's working

TypedSyntax/src/show.jl
charleskawczynski: I think I'd prefer that users can toggle this, so that they can specify the level of type information.

charleskawczynski:
> For the tests, https://github.com/JuliaDebug/Cthulhu.jl/blob/master/test/test_codeview.jl looks like a good place to put it and hopefully the other tests in that file will help you in writing one for this. E.g. I think you should be able to adapt...

Hm, my first attempt doesn't seem to be limiting the types:

```julia
#=
using Revise; include(joinpath("test", "test_depth_limited_type_printing.jl"))
=#
import Cthulhu

Base.@kwdef struct Nested{A,B}
    num::Int = 1
end
bar(x) = rand() > 0.5 ? x : Any[0][1]
mysum(x) = sum(y-> bar(x.num), 1:5; init=0)
nest_val(na, nb, ::Val{1}) = Nested{na, nb}()
nest_val(na, nb, ::Val{n}) where {n} = nest_val(Nested{na, nb}, Nested{na, nb}, Val(n-1))
nest_val(na, nb, n::Int) = nest_val(na, nb, Val(n))
nest_val(n) = nest_val(1, 1, n)
const NV = nest_val(5)

# f = nest_val(5)
# a = Any[f];
# mysum(a[1]) # make sure it runs
# Cthulhu.@descend mysum(a[1]) # navigate to sum -> sum, and Nested will be there
using Test
include("setup.jl")
@testset "hide type-stable statements" begin
    let # optimize code
        # f = nest_val(5)
        # a = Any[f];
        # mysum(a[1]) # make sure it runs
        # Cthulhu.@descend mysum(a[1]) # navigate to sum -> sum, and Nested will be there
        (; src, infos, mi, rt, exct, effects, slottypes) = @eval Module() begin
            $cthulhu_info($mysum, ($(typeof(NV)),))
        end;
        function prints(; kwargs...)
            io = IOBuffer()
            ioc = IOContext(io, :maxtypedepth => Cthulhu.CONFIG.type_depth_limit)
            Cthulhu.cthulhu_typed(ioc, :none, src, rt, exct, effects, mi; kwargs...)
            return String(take!(io))
        end;

        let # by default, should print every statement
            s = prints()
            println(s)
            # @test occursin("::Nested{Nested{…}, Nested{…}}", s)
        end
    end
end
```

Any pointers?

Zentrik:
I don't see anything obvious but I'm very busy right now so I'm not going to be able to get back to this pr for a few weeks.
Thanks for your work so far, especially in addressing my comments.

charleskawczynski:
#566 makes this package virtually unusable for us in some applications. Can someone please take a look at this? This PR does fix the issue, I think the test just needs someone's attention who is familiar with the style of the existing tests (and perhaps for someone to try with VScode integration).

charleskawczynski:
Actually, it looks like some of the types are printed using `InteractiveTools.warntype_type_printer`, so we need to fix this in julialang. At which point, we might as well and try to fix https://github.com/julialang/julia/issues/55952 first, and spread its use to InteractiveUtils, JET and here. I think that's probably a better course of action..

Zentrik (commented):
Sorry, for dropping the ball on this. I had a brief look and this lgtm

charleskawczynski:
> Sorry, for dropping the ball on this. I had a brief look and this lgtm

It’s alright. Honestly, I think we need to fix https://github.com/julialang/julia/issues/55952 and formulate the functions to return a depth-limited string (for readability), since one (or more) of the types is printed with IntwractiveUtils

serenity4:
I'd prefer if we had a UI option to increase/decrease the type depth limit (or directly set a number). Most of the time, it is preferable to get all of the information (even if harder to read) rather than incomplete information with no way to retrieve it; the exception being of course the case where it completely makes it unusable due to heavily nested types spamming the terminal. I am notably thinking of cases where perhaps one or two expressions may have heavily nested types in a function, and these being of interest (e.g. to detect non-concrete types); that'd be frustrating to only see part of it with really no way to dig deeper while that is the main point of Cthulhu.

Then if we have an UI option, we could set the default type depth limit to something higher, perhaps 6 or 7, to keep roughly the same behavior as currently for most cases, while avoiding non-negligible performance/readability issues for very heavily nested types. What do you think?

charleskawczynski:
> I'd prefer if we had a UI option to increase/decrease the type depth limit (or directly set a number). Most of the time, it is preferable to get all of the information (even if harder to read) rather than incomplete information with no way to retrieve it; the exception being of course the case where it completely makes it unusable due to heavily nested types spamming the terminal. I am notably thinking of cases where perhaps one or two expressions may have heavily nested types in a function, and these being of interest (e.g. to detect non-concrete types); that'd be frustrating to only see part of it with really no way to dig deeper while that is the main point of Cthulhu:
> 
> ```julia
> julia> function f(x)
>          x[][] + 1
>        end
> f (generic function with 1 method)
> 
> julia> @descend f(Ref(Ref(5)))
> f(x) @ Main ~/.julia/dev/Cthulhu/script.jl:4
> 4 function f(x::Base.RefValue{Base.RefValue{…}})::Int64
> 5   (x::Base.RefValue{Base.RefValue{…}}[]::Base.RefValue{Int64}[]::Int64 + 1)::Int64
> 6 end
> Select a call to descend into or ↩ to ascend. [q]uit. [b]ookmark.
> Toggles: [w]arn, [h]ide type-stable statements, [t]ype annotations, [s]yntax highlight for Source/LLVM/Native, [j]ump to source always, [v]scode: inlay types, [V]scode: diagnostics.
> Show: [S]ource code, [A]ST, [T]yped code, [L]LVM IR, [N]ative code
> Actions: [E]dit source code, [R]evise and redisplay
>  • x::Base.RefValue{Base.RefValue{…}}[]
>    x::Base.RefValue{Base.RefValue{…}}[]::Base.RefValue{Int64}[]
>    x::Base.RefValue{Base.RefValue{…}}[]::Base.RefValue{Int64}[]::Int64 + 1
>    ↩
> ```
> 
> Then if we have a UI option, we could set the default type depth limit to something higher, perhaps 6 or 7, to keep roughly the same behavior as currently for most cases, while avoiding non-negligible performance/readability issues for very heavily nested types. What do you think?

Yeah, that would be nice. I'm not really familiar with how to UI works, so I might need pointers for that. This is why I just configured things so that there is a global module setting that users can change.

Unfortunately, as I mentioned, some of the types are printed in `InteractiveTools.warntype_type_printer`-- and this is printed with syntax highlighting, so we can't easily/elegantly print this to a buffer and apply depth-limited type printing from within Cthulhu. i.e., we need a fix in julialang.

PR #585, opened 2024-08-04
Stop unwrapping types while mapping

timholy:
This started out as just an attempt to get the remaining tests working on Julia 1.11. However, one of the failures was a case where we wanted to preserve a `Core.Const` wrapper, and as I dug in it became increasingly clear that it was a little risky to alter the type assigned by inference before mapping.

This PR preserves (or, intends to preserve) the exact types assigned by inference. It then becomes the `show` component's job to strip `Core.Const` and similar wrapper before displaying types.

@Zentrik, now the remaining failures seem to be in the vscode portions. Any chance you could take a look at this? I'm less familiar with that code.

### DIFF
--- a/TypedSyntax/src/node.jl
+++ b/TypedSyntax/src/node.jl
@@ -191,12 +191,12 @@ function map_signature!(sig::TypedSyntaxNode, slotnames::Vector{Symbol}, slottyp
         kwdivider = 1
         if havekws && slotnames[1] !== Symbol("#self#")
             kwdivider = findfirst(1:length(slotnames)) do i
-                slotnames[i] == Symbol("") && unwrapinternal(slottypes[i]) <: Function  # this should be the parent function as an argument
+                slotnames[i] == Symbol("") && isa(unwrapinternal(slottypes[i]), Function)  # this should be the parent function as an argument
             end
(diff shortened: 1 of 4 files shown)


timholy:
Oh wait, it ended up being really simple. The question is whether one things `0::Core.Const(0)` is a regression compared to `0::Int`?

Zentrik:
Seems like an improvement to me given it's much more specific than Int.

timholy:
Cool. I'll look into that "exhaustive.jl" failure, and if/when fixed we should have TypedSyntax tests passing on 1.10 and 1.11 (but not nightly).

Given that inference is allowed to change between releases, I don't think any of these changes count as breaking? (Cthulhu is also the only dependent of TypedSyntax listed on juliahub.)

timholy:
Let's see if CodeTracking PR130 fixed the `exhaustive.jl` test

timholy:
TypedSyntax is now passing on 1.11. Since this is a fairly significant change and I'm not 100% confident in our test coverage (as good as it is), I'll play with it interactively a little before merging.

timholy:
This seems to be working reasonably well and without egregious mistakes, so let's merge this.

PR #587, opened 2024-08-14
consistency of `[@]descend_code_[typed|warntype]` with InteractiveUtils

simeonschaub:
I would expect these macros to behave like their InteractiveUtils
counterparts and show the annotated IR instead of annotated source. The
behavior of `[@]descend` remains the same but is now decoupled from
`descend_code_typed`.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -106,7 +106,7 @@ export ascend
     @descend_code_typed
 
 Evaluates the arguments to the function or macro call, determines their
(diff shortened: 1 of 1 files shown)


timholy (approved):
Good catch. I think one can call this a bugfix, but it could also be a breaking change 🤷‍♂️ .

simeonschaub:
Since it only really affects interactive use, I would agree that this should be fine in a minor release

vchuravy (approved):
Fine as a minor.

PR #588, opened 2024-08-15
Allow a `Tuple{typeof(f),...}` as root of ascend tree

timholy:
Required for https://github.com/aviatesk/JET.jl/pull/648#issuecomment-2280722393

### DIFF
--- a/src/callsite.jl
+++ b/src/callsite.jl
@@ -519,6 +519,7 @@ maybe_callsite(info::PureCallInfo, mi::MethodInstance) = mi.specTypes <: Tuple{m
 maybe_callsite(info::RTCallInfo, mi::MethodInstance) = false
 
 function maybe_callsite(info::RTCallInfo, @nospecialize(tt::Type))
+    tt === Union{} && return false
     info.f === tt.parameters[1] && info.argtyps == tt.parameters[2:end]
 end
 function maybe_callsite(info::MICallInfo, @nospecialize(tt::Type))
--- a/src/callsite.jl
+++ b/src/callsite.jl
@@ -506,6 +506,7 @@ end
 
 # maybe_callsite for higher-level inputs
 maybe_callsite(cs::Callsite, callee::MethodInstance) = maybe_callsite(cs.info, callee)
+maybe_callsite(cs::Callsite, @nospecialize(tt::Type)) = maybe_callsite(cs.info, tt)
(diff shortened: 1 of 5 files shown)


src/callsite.jl
```diff
@@ -517,4 +518,13 @@ end
 maybe_callsite(info::PureCallInfo, mi::MethodInstance) = mi.specTypes <: Tuple{mapany(Core.Typeof ∘ unwrapconst, info.argtypes)...}
 maybe_callsite(info::RTCallInfo, mi::MethodInstance) = false
 
+function maybe_callsite(info::RTCallInfo, @nospecialize(tt::Type))
+    info.f === tt.parameters[1] && info.argtyps == tt.parameters[2:end]
+end
+function maybe_callsite(info::MICallInfo, @nospecialize(tt::Type))
+    info.mi.specTypes <: tt
+end
```
timholy, line 526: I'm uncertain whether I've covered all the relevant cases.

timholy:
I've been using this for a couple of hours and it's been great, so I think it's baked well enough to release.

src/callsite.jl
```diff
@@ -517,4 +518,19 @@ end
 maybe_callsite(info::PureCallInfo, mi::MethodInstance) = mi.specTypes <: Tuple{mapany(Core.Typeof ∘ unwrapconst, info.argtypes)...}
 maybe_callsite(info::RTCallInfo, mi::MethodInstance) = false
 
+function maybe_callsite(info::RTCallInfo, @nospecialize(tt::Type))
+    isa(tt, Union) && return maybe_callsite(info, tt.a) || maybe_callsite(info, tt.b)
+    isa(tt, DataType) || return false
+    typeof(info.f) === tt.parameters[1] || return false
+    for (a, b) in zip(info.argtyps, tt.parameters[2:end])
+        a === b || return false
```
aviatesk, line 526: When iterating over a `zip` created from a `Vector{Any}`, objects of type `Tuple{<:Any,<:Any}` are created, which can sometimes lead to unnecessary specializations, especially when the vector contains `Type` objects.
Since we're only using `===`, it shouldn't be a problem for this case, but it's something to be mindful of when working on `Core.Compiler` and other external abstract interpreters in the future.

timholy: Yeah, I figured this case would be OK. I was leaning on the "self check" tests too, I admit!

PR #613, opened 2024-11-14
use `Compiler` stdlib if available

aviatesk:
This requires JuliaLang/julia#56553 to be merged, and the new Compiler.jl stdlib to be registered in General beforehand.


aviatesk:
Ok, this package has been fully updated to the Compiler.jl separation. Ready to go.

PR #619, opened 2025-02-19
make Cthulhu use Compiler.jl as stdlib optionally when it's loaded

aviatesk:
Using the Compiler stdlib does make things a bit tricky because we have
to consider Compiler.jl's compatibility not just with the package itself,
but also with the runtime system. In the past, we only had to worry
about the former, so issues didn't come up as often (since using
`Core.Compiler` guarantees that the compiler is compatible with the
runtime system always).

This commit changes Cthulhu back to using `Base.Compiler` until the user
specifically loads the `Compiler.jl` standard library.
This should reduce how often Cthulhu breaks (in normal use).

When an user loads Compiler explicitly, then Cthulhu now load the
CthulhuCompilerExt extension module, and wraps the implementations of
exported APIs.

### DIFF
--- a/ext/CthulhuCompilerExt.jl
+++ b/ext/CthulhuCompilerExt.jl
@@ -6,6 +6,7 @@ module CthulhuCompilerExt
     using Cthulhu: Cthulhu
 
     function __init__()
+        CC.activate!(; reflection=true)
         Cthulhu.CTHULHU_MODULE[] = @__MODULE__
     end
 
--- a/ext/CthulhuCompilerExt.jl
+++ b/ext/CthulhuCompilerExt.jl
@@ -6,7 +6,6 @@ module CthulhuCompilerExt
     using Cthulhu: Cthulhu
 
     function __init__()
-        CC.activate!(; reflection=true)
         Cthulhu.CTHULHU_MODULE[] = @__MODULE__
     end
 
--- a/ext/CthulhuCompilerExt.jl
+++ b/ext/CthulhuCompilerExt.jl
@@ -1,0 +1,15 @@
+module CthulhuCompilerExt
+
+@static if VERSION ≥ v"1.12.0-DEV.1581"
+    using Compiler: Compiler as CC
+    using Compiler.IRShow: IRShow
+    using Cthulhu: Cthulhu
+
+    function __init__()
+        Cthulhu.CTHULHU_MODULE[] = @__MODULE__
+    end
+
+    include("../src/CthulhuBase.jl")
+end
+
+end
--- a/Project.toml
+++ b/Project.toml
@@ -24,7 +24,7 @@ CthulhuCompilerExt = "Compiler"
 
 [compat]
 CodeTracking = "0.5, 1"
-Compiler = "0.0.3"
+Compiler = "0.0.3, 1.12.0"
 FoldingTrees = "1"
 InteractiveUtils = "1.9"
 JuliaSyntax = "0.4"
--- a/Project.toml
+++ b/Project.toml
@@ -24,7 +24,7 @@ CthulhuCompilerExt = "Compiler"
 
 [compat]
 CodeTracking = "0.5, 1"
-Compiler = "0.0.3, 1.12.0"
+Compiler = "0, 1"
 FoldingTrees = "1"
 InteractiveUtils = "1.9"
 JuliaSyntax = "0.4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,22 @@
+authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
 name = "Cthulhu"
 uuid = "f68482b8-f384-11e8-15f7-abe071a5a75f"
 version = "2.16.4"
-authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
+
+[compat]
+CodeTracking = "0.5, 1"
+Compiler = "0, 1"
+FoldingTrees = "1"
+InteractiveUtils = "1.9"
+JuliaSyntax = "0.4"
+PrecompileTools = "1"
+Preferences = "1"
+REPL = "1.9"
+TypedSyntax = "1.3.0"
+UUIDs = "1.9"
+Unicode = "1.9"
+WidthLimitedIO = "1"
+julia = "1.10"
 
 [deps]
 CodeTracking = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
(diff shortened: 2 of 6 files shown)


aviatesk:
Alternatively, it might be simpler to keep Compiler.jl as a regular (not used by default) dependency in Project.toml and switch `Cthulhu.CC` to `Base.Compiler` or the stdlib `Compiler` based on the settings in Preferences.jl. Any thought on this?

vchuravy:
> Alternatively, it might be simpler to keep Compiler.jl as a regular (not used by default) dependency in Project.toml and switch Cthulhu.CC to Base.Compiler or the stdlib Compiler based on the settings in Preferences.jl. Any thought on this?

I would do it the opposite? Use `Compiler` as default and `Base.Compiler` when a preference asks for it? 
I.e. the normal case of using Cthulhu when not using `julia#main` should be the priority.

aviatesk:
> > Alternatively, it might be simpler to keep Compiler.jl as a regular (not used by default) dependency in Project.toml and switch Cthulhu.CC to Base.Compiler or the stdlib Compiler based on the settings in Preferences.jl. Any thought on this?
> 
> I would do it the opposite? Use `Compiler` as default and `Base.Compiler` when a preference asks for it? I.e. the normal case of using Cthulhu when not using `julia#main` should be the priority.

The problem is we don't have clear rules for releasing the Compiler.jl stdlib. So we often find situations where regular users not using `julia#master` (like those on v1.12 nightly) end up using a Compiler.jl stdlib incompatible with v1.12 when they use Cthulhu, since the compiler's stdlib gets released at random times, based on the `#master` branch code at that moment, without considering e.g. backports.

So I think the default should still be `Base.Compiler`.

vchuravy:
Shouldn't the Compiler lib then declare itself incompatible with 1.12?

aviatesk:
Yeah, managing Compiler.jl versions is a general problem.

We need to discuss and decide how to handle compatible Julia runtime versions, or how to release Compiler.jl versions that support older runtimes.

This pull request is a temporary fix until we set those rules.

Even if this is a temporal solution, we need to decide at least one of the following:
- Which compiler implementation to load by load, `Base.Compiler` or the Compiler stdlib.
- How to switch Compiler implementations: using Preferences.jl, or using a package extension that hooks loading Compiler.jl.

ext/CthulhuCompilerExt.jl
```diff
@@ -0,0 +1,16 @@
+module CthulhuCompilerExt
+
+@static if VERSION ≥ v"1.12.0-DEV.1581"
+    using Compiler: Compiler as CC
+    using Compiler.IRShow: IRShow
+    using Cthulhu: Cthulhu
+
+    function __init__()
```
Keno, line 8: I don't think this should happen automatically - instead it should probably switch on the base version of the compiler.

KristofferC:
Bump

ext/CthulhuCompilerExt.jl
aviatesk: I'm not sure what you meant by that. Does that mean the Compiler stdlib is only loaded when the `Base.Compiler` and the actual Compiler stdlib implementation are different? Basically, only if someone has made local changes to the Compiler implementation.

aviatesk:
@Keno Does this PR look reasonable to you?

serenity4:
From what I can see, SnoopCompile only interacts with `Base.Compiler`, never with the Compiler stdlib, so there might be an easy fix: switch all uses of `instance` inside Cthulhu to use `Cthulhu.instance` and ignore methods for `CthulhuCompilerExt`.

To have a better solution to the general problem, we'd want to only re-evaluate parts that directly use Compiler in the `CthulhuCompilerExt` extension (instead of re-including most of the package), therefore only define `instance` in Cthulhu and not in the extension, which #662 should make much easier to do.

timholy:
The problem is this:
```julia
tim@kenzo:~/.julia/dev/SnoopCompile$ julia +nightly -q
julia> filter(collect(keys(Base.loaded_modules))) do id
                  id.name == "Compiler"
              end
1-element Vector{Base.PkgId}:
 Compiler [807dbc54-b67e-4c79-8afb-eafe4df6f2e1]
```
and now cross-check that UUID against [this](https://github.com/JuliaDebug/Cthulhu.jl/blob/8a40464cc1051b82a74e65cae3b799331f7f0f35/Project.toml#L35).

This only affects interactive sessions:
```sh
tim@kenzo:~/.julia/dev/SnoopCompile$ julia +nightly -q -e 'filter(collect(keys(Base.loaded_modules))) do id id.name == "Compiler" end'
tim@kenzo:~/.julia/dev/SnoopCompile$
```
Because precompilation is non-interactive, it means that during precompilation, SnoopCompile never "sees" CthulhuCompilerExt.

However, using `ascend` in SnoopCompile is *always* interactive. Thus:
- during precompilation, SnoopCompile can't add the `instance` method(s) to CthulhuCompilerExt
- during usage, only CthulhuCompilerExt matters

Annoyingly, even disabling precompilation of my extension seems not to work:
```julia
__precompile__(false)

module CthulhuExt
    using Cthulhu: Cthulhu
    using Core: MethodInstance
    using SnoopCompile: InstanceNode, TriggerNode, Suggested, InferenceTrigger, countchildren, callingframe, callerinstance

    const C = Cthulhu.CTHULHU_MODULE[]
    @show C
    ...
end
```
still produces
```julia
julia> using SnoopCompile, Cthulhu
[ Info: Precompiling CthulhuExt [54ad8b97-6788-5604-ab44-4ecd179750f8] (cache misses: include_dependency fsize change (6), wrong dep version loaded (14))
┌ Info: Skipping precompilation due to precompilable error. Importing CthulhuExt [54ad8b97-6788-5604-ab44-4ecd179750f8].
└   exception = Error when precompiling module, potentially caused by a __precompile__(false) declaration in the module.
C = Cthulhu
```

CC @KristofferC in case he has any ideas.

timholy:
Amusingly, if I tried to add those methods in `SnoopCompile/ext/CthulhuExt`'s `__init__()` to e.g., `C.instance`, I get different answers depending on load order:
- `using SnoopCompile, Cthulhu` causes `SnoopCompile/ext/CthulhuExt.__init__()` to run first, so that `C = Cthulhu` and thus adding the methods to `Cthulhu`.
- `using Cthulhu, SnoopCompile` runs `CthulhuCompilerExt.__init__()` first, so that `C = CthulhuCompilerExt` and adding the methods to the correct module and letting my modified version of SnoopCompile/test/extensions/cthulhu.jl pass.

But making this dependent on load order is horrible.

PR #621, opened 2025-02-19
adjust Cthulhu to JuliaLang/julia#56248

aviatesk:


serenity4:
@aviatesk could you release a new version containing the changes in this PR? I'd like to reuse `set_cthulhu_source!(result::InferenceResult)` in https://github.com/JuliaDiff/Diffractor.jl/pull/300.

This function is largely internal, but Diffractor seems to anyways rely on Cthulhu internals in lots of places, so it seems more convenient rather than using `result.src = create_cthulhu_source(result.src, result.ipo_effects)`.

aviatesk:
Released version 2.16.3 for now.

PR #623, opened 2025-02-28
Adapt to `finishinfer!` signature change

serenity4:
`finishinfer!` changed its signature with https://github.com/JuliaLang/julia/pull/57545, adding a `cycleid::Int` argument.

### DIFF
--- a/src/interpreter.jl
+++ b/src/interpreter.jl
@@ -105,7 +105,7 @@ function InferredSource(state::InferenceState)
         exct)
 end
 
-@static if VERSION ≥ v"1.13.0-DEV.126"
+@static if VERSION ≥ v"1.13.0-DEV.126" || VERSION ≥ v"1.12.0-alpha1"
 function cthulhu_finish(@specialize(finishfunc), state::InferenceState, interp::CthulhuInterpreter, cycleid::Int)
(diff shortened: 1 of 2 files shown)


src/interpreter.jl
```diff
@@ -105,12 +105,22 @@ function InferredSource(state::InferenceState)
         exct)
 end
 
+@static if VERSION ≥ v"1.13.0-DEV.126" || v"1.12.0-alpha1" ≤ VERSION < v"1.13"
```
serenity4, line 108: The fix has been marked as a 1.12 backport so this check allows us to not break nightly versions for `v"1.13.0-DEV.0" < VERSION < v"1.13.0-DEV.126"`. On the other hand it's a bit complex, I'm not sure how we usually deal with backports in this situation.

Happy to change that (and the other check below) to `VERSION ≥ v"1.12.0-alpha1"` if the complex version management is not deemed worth it.

aviatesk: We don't have a solid rule for this sort of things, but I think I'd prefer to having a simpler versioning here, so `VERSION ≥ v"1.12.0-alpha1"` would be best. We generally don't account those who use outdated nightly version.

aviatesk (approved):
Thanks!

PR #624, opened 2025-03-04
Fix test suite

serenity4:
This should fix test failures for nightly, so it is easier to test & approve changes for PRs.

### DIFF
--- a/TypedSyntax/test/runtests.jl
+++ b/TypedSyntax/test/runtests.jl
@@ -3,7 +3,7 @@ using TypedSyntax: TypedSyntax, TypedSyntaxNode
 using Dates, InteractiveUtils, Test
 using Core.Compiler: widenconst
 
-has_typ(node, @nospecialize(Ts::Tuple)) = any(has_typ(node, T) for T in Ts)
+has_typ(node, @nospecialize(Ts::Tuple)) = any(has_typ(node, Ts[i]) for i in 1:length(Ts))
 has_typ(node, @nospecialize(T)) = node.typ === T || !isnothing(node.typ) && widenconst(node.typ) === T
 has_name_typ(node, name::Symbol, @nospecialize(T)) = kind(node) == K"Identifier" && node.val === name && has_typ(node, T)
 has_name_notyp(node, name::Symbol) = has_name_typ(node, name, nothing)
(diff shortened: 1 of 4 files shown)


serenity4:
Ah, well, it will actually require #619 first.

TypedSyntax/test/runtests.jl
```diff
@@ -1,9 +1,11 @@
 using JuliaSyntax: JuliaSyntax, SyntaxNode, children, child, sourcetext, kind, @K_str
 using TypedSyntax: TypedSyntax, TypedSyntaxNode
 using Dates, InteractiveUtils, Test
+using Core.Compiler: widenconst
 
-has_name_typ(node, name::Symbol, @nospecialize(Ts::Tuple)) = kind(node) == K"Identifier" && node.val === name && node.typ in Ts
-has_name_typ(node, name::Symbol, @nospecialize(T)) = kind(node) == K"Identifier" && node.val === name && node.typ === T
+has_typ(node, @nospecialize(Ts::Tuple)) = any(has_typ(node, T) for T in Ts)
+has_typ(node, @nospecialize(T)) = node.typ === T || !isnothing(node.typ) && widenconst(node.typ) === T
+has_name_typ(node, name::Symbol, @nospecialize(T)) = kind(node) == K"Identifier" && node.val === name && has_typ(node, T)
```
serenity4, line 8: Not 100% sure about these changes. Essentially, what I have been assuming is that it doesn't matter whether the node type is a `Const` or not, so long as the type of the value is correct, if we test it against a non-`Const` type.

aviatesk: Yeah, this change looks reasonable.

TypedSyntax/test/runtests.jl
```diff
@@ -1,9 +1,11 @@
 using JuliaSyntax: JuliaSyntax, SyntaxNode, children, child, sourcetext, kind, @K_str
 using TypedSyntax: TypedSyntax, TypedSyntaxNode
 using Dates, InteractiveUtils, Test
+using Core.Compiler: widenconst
 
-has_name_typ(node, name::Symbol, @nospecialize(Ts::Tuple)) = kind(node) == K"Identifier" && node.val === name && node.typ in Ts
-has_name_typ(node, name::Symbol, @nospecialize(T)) = kind(node) == K"Identifier" && node.val === name && node.typ === T
+has_typ(node, @nospecialize(Ts::Tuple)) = any(has_typ(node, T) for T in Ts)
```
aviatesk, line 6: Just a minor nit: It would be usually better to avoid iterating over tuple whose type isn't statically known, since it can result in less profitable code generation.
```suggestion
has_typ(node, @nospecialize(Ts::Tuple)) = any(has_typ(node, Ts[i]) for i in 1:length(Ts))
```

aviatesk (approved):
Thanks for this update. While I'm working on #619, this PR doesn't need to wait for it since it already passes all tests on the stable version.

PR #614, opened 2024-11-21
wip: make Cthulhu cache `CodeInstance`-based

aviatesk:
Although these code paths don’t seem to be used very often.

PR #628, opened 2025-03-18
Make Cthulhu cache `CodeInstance`-based, continued

serenity4:
Continues the work started in https://github.com/JuliaDebug/Cthulhu.jl/pull/614.

Tests are passing locally, this is ready for review.

### DIFF
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -507,10 +507,9 @@ function _descend(term::AbstractTerminal, interp::AbstractInterpreter, curs::Abs
     display_CI = true
     view_cmd = cthulhu_typed
     iostream = term.out_stream::IO
-    function additional_descend(new_mi::MethodInstance)
+    function additional_descend(new_ci::CodeInstance)
         new_interp = CthulhuInterpreter(interp)
-        do_typeinf!(new_interp, new_mi)
-        _descend(term, new_interp, new_mi;
+        _descend(term, new_interp, new_ci;
                  debuginfo, optimize, interruptexc, iswarn, hide_type_stable, remarks,
                  with_effects, exception_type,
                  inline_cost, type_annotations, annotate_source,
@@ -538,20 +537,21 @@ function _descend(term::AbstractTerminal, interp::AbstractInterpreter, curs::Abs
                         # but make something up that looks plausible.
                         callsites = Callsite[]
                         if display_CI
-                            exct = @static VERSION ≥ v"1.11.0-DEV.945" ? codeinst.exct : nothing
+                            exct = @static VERSION ≥ v"1.11.0-DEV.945" ? codeinst.exctype : nothing
                             callsite = Callsite(-1, EdgeCallInfo(codeinst, codeinst.rettype, get_effects(codeinst), exct), :invoke)
                             println(iostream)
                             println(iostream, "│ ─ $callsite")
                             println(iostream, "│  return ", Const(codeinst.rettype_const))
                             println(iostream)
                         end
+                        mi = codeinst.def
                         @goto show_menu
                     else
                         @info """
                         Inference discarded the source for this call because of recursion:
                         Cthulhu nevertheless is trying to retrieve the source for further inspection.
                         """
-                        additional_descend(get_mi(curs))
+                        additional_descend(get_ci(curs))
                         break
                     end
                 end
@@ -575,7 +575,7 @@ function _descend(term::AbstractTerminal, interp::AbstractInterpreter, curs::Abs
         end
         pc2excts = nothing
         end
-        callsites, sourcenodes = find_callsites(interp, src, infos, mi, slottypes, optimize & !annotate_source, annotate_source, pc2excts)
+        callsites, sourcenodes = find_callsites(interp, src, infos, ci, slottypes, optimize & !annotate_source, annotate_source, pc2excts)
 
         if jump_always
             if isdefined(Main, :VSCodeServer) && Main.VSCodeServer isa Module && isdefined(Main.VSCodeServer, :openfile)
@@ -671,7 +671,7 @@ function _descend(term::AbstractTerminal, interp::AbstractInterpreter, curs::Abs
                 if sourcenode !== nothing
                     show_sub_callsites = let callsite=callsite
                         map(info.callinfos) do ci
-                            p = Base.unwrap_unionall(get_mi(ci).specTypes).parameters
+                            p = Base.unwrap_unionall(ci.def.specTypes).parameters
                             if isa(sourcenode, TypedSyntax.MaybeTypedSyntaxNode) && length(p) == length(JuliaSyntax.children(sourcenode)) + 1
                                 newnode = copy(sourcenode)
                                 for (i, child) in enumerate(JuliaSyntax.children(newnode))
@@ -705,7 +705,7 @@ function _descend(term::AbstractTerminal, interp::AbstractInterpreter, curs::Abs
                 Inference didn't analyze this call because it is a dynamic call:
                 Cthulhu nevertheless is trying to descend into it for further inspection.
                 """
-                additional_descend(get_mi(info)::MethodInstance)
+                additional_descend(get_ci(info)::CodeInstance)
                 continue
             elseif info isa RTCallInfo
                 @info """
--- a/src/Cthulhu.jl
+++ b/src/Cthulhu.jl
@@ -347,26 +347,39 @@ end
 
 get_effects(codeinst::CodeInstance) = CC.decode_effects(codeinst.ipo_purity_bits)
 get_effects(codeinst::CodeInfo) = CC.decode_effects(codeinst.purity)
-get_effects(src::InferredSource) = src.effects
 get_effects(result::InferenceResult) = result.ipo_effects
 get_effects(result::CC.ConstPropResult) = get_effects(result.result)
 get_effects(result::CC.ConcreteResult) = result.effects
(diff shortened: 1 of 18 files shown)


src/interpreter.jl
```diff
@@ -259,7 +288,7 @@ end
 
 function CC.IRInterpretationState(interp::CthulhuInterpreter,
     code::CodeInstance, mi::MethodInstance, argtypes::Vector{Any}, world::UInt)
-    inferred = @atomic :monotonic code.inferred
+    inferred = code.inferred
```
serenity4, line 291: That was a change from your original PR, is `CodeInstance` guaranteed to be "owned" by the current compilation? Would that be what `engine_reserve`/`engine_reject` do?

aviatesk: If you declare a field with `@atomic`, the necessary fence will be automatically added in its `getfield` operation. So, you don't need the `@atomic` annotation at the call site. We can leave it there for clarity, though.

serenity4: Ah yes, I see, it's already taken care of in the struct declaration. No need to think about it there then!

src/Cthulhu.jl
```diff
@@ -845,57 +842,55 @@ end
 
 function do_typeinf!(interp::AbstractInterpreter, mi::MethodInstance)
     result = InferenceResult(mi)
-    @static if isdefined(CC, :engine_reserve)
-        ci = CC.engine_reserve(interp, mi)
-        result.ci = ci
-    end
+    ci = CC.engine_reserve(interp, mi)
```
aviatesk, line 845: This PR requires us to bump the minimum compatible Julia version to 1.11, requiring Cthulhu's minor version to be bumped jointly.

serenity4: Oh yes, I think these line edits were made in your original PR. Do you think that this PR is fundamentally incompatible with v1.11 (e.g. because code instances are not available as much), or can we do with keeping support for v1.11 while throwing out the `MethodInstance`-based caching?

aviatesk (commented):
https://github.com/serenity4/cthulhu.jl/blob/d3ff8213541262ec36f08052b15844a44fefe4b1/src/codeview.jl#L336-L395
Do you think we could refactor this part to use `CodeInstance` for lookups instead of `MethodInstance`? It looks like it might let us get rid of `get_mi` completely.

But this `add_callsites!` function is only used by an experimental feature in VSCode. So, if it seems too hard or would need a bigger refactoring, it's totally fine to leave it as it is and half-supported.

serenity4:
Good point, I'll take a look :+1:

aviatesk:
When I handed this PR over to you, I felt that completing it would require a more extensive refactoring than the commits you added (the refactoring to split Cthulhu into the UI part and the `AbstractInterpreter` integration part, as we discussed before). But I can't remember what piece of code made me feel like that...

That said, looking at this code, it seems like most Cthulhu features are probably supported on this PR, so I think it's okay to proceed as is for now.
We'll probably do that extensive refactoring eventually anyway.

src/Cthulhu.jl
serenity4: Feel free to create the branch with a name of your choosing, and I'll redirect the PR to it.

aviatesk: ok, I created one named `2.17`.

serenity4:
> https://github.com/serenity4/cthulhu.jl/blob/d3ff8213541262ec36f08052b15844a44fefe4b1/src/codeview.jl#L336-L395
Do you think we could refactor this part to use CodeInstance for lookups instead of MethodInstance? It looks like it might let us get rid of `get_mi` completely.

`TaskCallInfo` (relevant in context of task-spawning callsites) currently wraps an `EdgeCallInfo` constructed with a `MethodInstance`, by manually materializing the `MethodInstance` that corresponds to the `start` function of the task. It is not trivial to get a `CodeInstance` for it, because the runtime accepts a generic function argument that will be resolved dynamically. Would it nonetheless be reasonable to run inference on the `MethodInstance` to get a `CodeInstance` back? I see that [we already run type inference when descending into it](https://github.com/JuliaDebug/Cthulhu.jl/blob/bfbe6456b833791080b6ab274c4871e043dabcfa/src/Cthulhu.jl#L702-L707), the difference would be that we now do that eagerly when creating the `TaskCallInfo`.

Other than that, I don't see major obstacles to use `CodeInstance`s for the cache-related parts in `add_callsites!`. I had a failure for `getproperty(::Module, ::Symbol)`, where unoptimized source does not appear to be available in lookup, which I'll look into. We also still use `MethodInstance`s for a few things (notably, `TypedSyntax`-related parts), which I am thinking might not gain much by using `CodeInstance`s, so the change does not have to be very intrusive since we can always extract one from the `CodeInstance`.

aviatesk:
> Would it nonetheless be reasonable to run inference on the `MethodInstance` to get a `CodeInstance` back?

Yeah, I think it's totally fine to change it that way. Eager inference might add a small performance hit to code with task constructs, but Cthulhu's performance doesn't need to be super optimized anyway. Whatever's easiest to deal with is good.

> I haven't managed to identify the exact cause beyond the fact that `finishinfer!` is never called on it, looks like inference does not run for this call

This issue sounds a bit serious. Does this only happen with inference performed from `add_callsites!`? I just looked back at the current implementation of abstractinterpretation.jl, and it seems like there is no longer a guarantee that `finishinfer!` is called for a frame when it is created from `const_prop_call`.

serenity4:
My mistake, `finishinfer!` is actually called; even in `const_prop_call`, inference goes through `typeinf(interp, frame)` which ensures that once frames are inferred they go through `finishinfer!`.

The issue is apparently that, in the cache, we store a `result::InferenceResult`, and later try to get an `InferredSource` from this cache using `ci::CodeInstance` such that `ci === result.ci_as_edge`.

serenity4:
I can suggest having a separate data structure in `CthulhuInterpreter` that store a list of `InferenceResult`s to be resolved to their `ci_as_edge` `CodeInstance`s after top-level inference is finished.

aviatesk:
Does this issue only happen for `add_callsites` where we need to lookup `InferredSource` from `InferenceKey`, or is this a general issue? AFAIU `result.ci_as_edge` doesn't contain `InferredSource`, so holding them in the interpreter doesn't allow us to inspect their inferred sources in the Cthulhu cache format.

serenity4:
> Does this issue only happen for `add_callsites` where we need to lookup InferredSource from InferenceKey, or is this a general issue?

It only happens there as far as I am aware, yes. I use the following reproducer:
```julia
using Cthulhu
f() = Base.sum
@descend f() # then descend into `Base.sum`
```

> AFAIU result.ci_as_edge doesn't contain InferredSource, so holding them in the interpreter doesn't allow us to inspect their inferred sources in the Cthulhu cache format.

I implemented a proof of concept here, this will be hopefully better communicate what I meant: https://github.com/JuliaDebug/Cthulhu.jl/pull/628/commits/844e6f4575848cde5b93fdb07784ae150049eaeb

This commit successfully the case above, but I can revert it if you prefer another approach; that was mainly to explore whether that might work in practice.

aviatesk:
Okay, I understand the situation.

Currently, the Cthulhu cache uses `Union{CodeInstance, InferenceResult}` as keys. But `add_callsites!` currently only looks up using `CodeInstance` (or `MethodInstance` previously) as the key. Ideally, like in Cthulhu's main driver, we should use `InferenceResult` as the key for call sites that have been const-prop'ed. However, the function haven't been implemented in that way, so I'm not sure if it's even possible.

A more general refactoring idea would be to stop using `InferenceResult` in the Julia compiler and manage everything associated with `CodeInstance`. That way, Cthulhu's cache could also be simplified to use `CodeInstance` as the only key type.

However, this feature isn't Cthulhu's main focus, and it's experimental anyway, so I'll leave it up to you to decide how to proceed.

serenity4:
Thanks for the clarifications. Thinking about it, I don't really like the idea of adding a field to `CthulhuInterpreter` for the purpose of patching this very issue, and at the same time it appears to me that exploring const-proped callsites is not particularly useful for `add_callsites!`. I reverted the previous commit and went with https://github.com/JuliaDebug/Cthulhu.jl/pull/628/commits/8d2b4fb14e3b25012415b5a0e9bd96d17c96ace2 which should fix the issue in a minimally intrusive way. We can proceed with this fix if you think that it is reasonable.

aviatesk:
I merged JuliaDebug/Cthulhu.jl#619, and also updated the `2.17` branch. I'm afraid but could you please rebase this branch? We can go ahead and merge this and release 2.17 then.

aviatesk:
Yeah, that's the plan.

In v1.11, the inference cache uses `MethodInstance` as the key, so we can't do this refactoring.

Starting with v1.12, we want to move forward with development based on this refactoring.

After v1.12 is officially released, we'll stop supporting v1.11. The we'll merge the `v2.17` branch into `master` and continue development as usual.

serenity4:
Sounds good! I updated the project and CI versions to use 1.12+.

serenity4:
After this is merged to `2.17` I'll make another PR to get rid of the outdated version checks to clean up the code a bit.

aviatesk (approved):

aviatesk:
Thanks for moving this PR forward!

aviatesk:
> After this is merged to `2.17` I'll make another PR to get rid of the outdated version checks to clean up the code a bit.

That's great. Please carry on.

PR #630, opened 2025-03-26
Remove outdated version checks

serenity4:
As we are removing support for Julia versions < 1.12 starting from the 2.17 release, we can remove all version checks that may be optimized away assuming `VERSION >= v"1.12-"`.


serenity4:
Looking at the warnings, I suspect the IntegrationTest job may be failing because JET does not yet support 1.12. As the update shouldn't be too far out, I think we can keep it just to make sure the error remains the same (and is not triggered by a failed precompilation of CthulhuExt for example).

aviatesk (approved):
Thanks!

Issue #633, opened 2025-04-11
Precompilation error on nightly


jishnub:
```julia
  1 dependency had output during precompilation:
┌ Cthulhu
│  ┌ Error: Errorred while running the precompile workload, the package may or may not work but latency will be long
│  │   exeption =
│  │    FieldError: type OptimizationState has no field `result`, available fields: `linfo`, `src`, `optresult`, `stmt_info`, `mod`, `sptypes`, `slottypes`, `inlining`, `cfg`, `unreachable`, `bb_vartables`, `insert_coverage`
│  │    Stacktrace:
│  │      [1] getproperty
│  │        @ ./Base_compiler.jl:54 [inlined]
│  │      [2] create_cthulhu_source
│  │        @ ~/Dropbox/JuliaPackages/Cthulhu.jl/src/interpreter.jl:128 [inlined]
│  │      [3] set_cthulhu_source!
│  │        @ ~/Dropbox/JuliaPackages/Cthulhu.jl/src/interpreter.jl:142 [inlined]
│  │      [4] finish!
│  │        @ ~/Dropbox/JuliaPackages/Cthulhu.jl/src/interpreter.jl:153 [inlined]
│  │      [5] finish_nocycle(::Cthulhu.CthulhuInterpreter, frame::Compiler.InferenceState, time_before::UInt64)
│  │        @ Compiler ./../usr/share/julia/Compiler/src/typeinfer.jl:217
│  │      [6] typeinf(interp::Cthulhu.CthulhuInterpreter, frame::Compiler.InferenceState)
│  │        @ Compiler ./../usr/share/julia/Compiler/src/abstractinterpretation.jl:4435
│  │      [7] top-level scope
│  │        @ ~/Dropbox/JuliaPackages/Cthulhu.jl/src/CthulhuBase.jl:690
│  │      [8] top-level scope
│  │        @ ~/Dropbox/JuliaPackages/Cthulhu.jl/src/CthulhuBase.jl:702
│  │      [9] top-level scope
│  │        @ ~/Dropbox/JuliaPackages/Cthulhu.jl/src/CthulhuBase.jl:724
│  │     [10] top-level scope
│  │        @ ~/Dropbox/JuliaPackages/Cthulhu.jl/src/CthulhuBase.jl:722
│  │     [11] top-level scope
│  │        @ ~/Dropbox/JuliaPackages/Cthulhu.jl/src/CthulhuBase.jl:138
│  │     [12] top-level scope
[...]

PR #634, opened 2025-04-11
Adjust to `.result` -> `.optresult` change for `OptimizationState`

serenity4:
Fixes https://github.com/JuliaDebug/Cthulhu.jl/issues/633 for version 2.17.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
 name = "Cthulhu"
 uuid = "f68482b8-f384-11e8-15f7-abe071a5a75f"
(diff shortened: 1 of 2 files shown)


serenity4:
Besides the renaming, an additional fix was needed: we relied on the optimizer setting the `.inlining_cost` field of the source `CodeInfo`, while with https://github.com/JuliaLang/julia/pull/57979 this operation is more explicit and set only when required during `finish!`. Because we create the `OptimizedSource` before `finish!`, and we rely on its `.isinlineable` field to be set for `src_inlining_policy`, it seems we must compute the inlining cost ourselves before hand.

I'll look into whether we really need to hold this information in `OptimizedSource`, don't merge this just yet.

PR #635, opened 2025-04-11
Backport: Adjust to .result -> .optresult change for OptimizationState

serenity4:
Backports https://github.com/JuliaDebug/Cthulhu.jl/pull/634 to have the fix for https://github.com/JuliaDebug/Cthulhu.jl/issues/633 on `master`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
 name = "Cthulhu"
 uuid = "f68482b8-f384-11e8-15f7-abe071a5a75f"
-version = "2.16.5"
+version = "2.16.6"
 
 [compat]
 CodeTracking = "0.5, 1"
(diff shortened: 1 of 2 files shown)


aviatesk:
Since users on Julia 1.12 or later will be using the 2.17 series, don't we need to add compatibility for Julia 1.12+ code to the 2.16 series?

serenity4:
> don't we need to add compatibility for Julia 1.12+ code to the 2.16 series?

Or did you say we don't need to add the code to the 2.16 series? That's true, it's not the first time I am confused about having 2.17 as a separate branch but having it available to 1.12 still :melting_face:

aviatesk:
Yeah, this is the first time we're doing things this way, and it turns out to be a bit complicated.
How about we flip the current setup? We manage 2.17 in the `master` branch, and the 2.16 series in the `2.16` branch, since we'll be making a lot of changes to 2.17, but probably hardly any to 2.16.

serenity4:
Yes, that's a good idea.

aviatesk:
> Or did you say we don't need to add the code to the 2.16 series?

Yeah, basically we just do "bug fix" only for the 2.16 series. If some destructive change is backported to Julia 1.11 series, we may need to apply some big changes to the 2.16 series, but otherwise we won't add much changes to 2.16.

PR #636, opened 2025-04-11
add compatibility warning to README

aviatesk:


README.md
```diff
@@ -1,8 +1,21 @@
...
+> This package relies on internal APIs of the Julia base compiler.
+> As such, its behavior may change depending on your Julia version.
+> For maintenance reasons, the latest version of Cthulhu (typically managed in the `master`
+> branch) does not guarantee compatibility with older Julia versions.
+> Versions of Cthulhu compatible with earlier Julia releases may be maintained in dedicated branches[^old-version-branch-ex].
+>
+> Please refer to the `Project.toml` file for details on supported Julia versions.
+> Generally, Julia's package manager automatically selects a compatible Cthulhu version for
+> your Julia installation.
+> If you simply want to use Cthulhu, running `pkg> add Cthulhu` will install an appropriate version.
```
serenity4, line 15: Did you mean for these `>` to be here?

serenity4:
If we plan to only support the latest nightly versions (to avoid version checks for unreleased versions), we could also briefly recommend to check users have the latest Cthulhu when updating their nightly version.

README.md
aviatesk: Yea, 
```md
> [!admonition-header]
> ...
```
is the syntax for showing admonition on GitHub markdown.

serenity4: For the `*The slow descent into madness*` too? This one seems outside of the admonition block.

aviatesk: We didn't really need that part, but the sentence itself is more like a casual comment than documentation, so I just quoted it;)

serenity4 (approved):

PR #637, opened 2025-04-11
update JuliaSyntax to v1.0

aviatesk:
I mechanically converted the affected parts to use the new interface, but the tests are failing for some reason. Maybe because the parsing itself has been changed between JuliaSyntax 0.4 and 1.0, meaning we probably need to update some parts of the TypedSyntax itself.


aviatesk:
@timholy How would you like to proceed? This PR can upgrade the JS used by TS to the latest stable version v1, but some detailed logic would be broken by this (and I don't have time to investigate the internals of TS at this moment). It would be ideal if we could easily fix the current implementation, but if you don't have time, I would like to merge this PR with the broken test cases as they are, since not being able to use Cthulhu in environments that use JS@1 is quite inconvenient.

PR #638, opened 2025-04-11
Refactor: extend cache transforms to store `OptimizedSource`

serenity4:
It seems that using the `transform_result_for_cache` and `transform_result_for_local_cache` mechanisms, we don't need to override `finish!` anymore.

Therefore, we now have:
- `CC.finishinfer!` to store inferred, unoptimized sources (`InferredSource`)
- `CC.transform_result_for_cache` and `CC.transform_result_for_local_cache` to store inferred, optimized sources (`OptimizedSource`)

### DIFF
--- a/src/interpreter.jl
+++ b/src/interpreter.jl
@@ -132,6 +132,9 @@ function CC.transform_result_for_local_cache(interp::CthulhuInterpreter, result:
 end
 
 function CC.transform_result_for_cache(interp::CthulhuInterpreter, result::InferenceResult, edges::Core.SimpleVector)
+    if isa(result.src, CodeInfo)
+        result.src.edges = edges
+    end
     result.src = CC.transform_result_for_local_cache(interp, result)
 end
 
--- a/src/interpreter.jl
+++ b/src/interpreter.jl
@@ -107,8 +107,8 @@ function InferredSource(state::InferenceState)
         exct)
 end
 
-function cthulhu_finish(@specialize(finishfunc), state::InferenceState, interp::CthulhuInterpreter, cycleid::Int)
-    res = @invoke finishfunc(state::InferenceState, interp::AbstractInterpreter, cycleid::Int)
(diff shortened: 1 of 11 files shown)


src/interpreter.jl
```diff
@@ -107,38 +107,35 @@ function InferredSource(state::InferenceState)
...
-
-CC.finishinfer!(state::InferenceState, interp::CthulhuInterpreter, cycleid::Int) = cthulhu_finish(CC.finishinfer!, state, interp, cycleid)
-function CC.finish!(interp::CthulhuInterpreter, caller::InferenceState, validation_world::UInt, time_before::UInt64)
-    set_cthulhu_source!(caller.result)
-    return @invoke CC.finish!(interp::AbstractInterpreter, caller::InferenceState, validation_world::UInt, time_before::UInt64)
+function CC.transform_result_for_cache(interp::CthulhuInterpreter, result::InferenceResult, edges::Core.SimpleVector)
+    if isa(result.src, CodeInfo)
+        result.src.edges = edges
+    end
+    result.src = CC.transform_result_for_local_cache(interp, result)
```
serenity4, line 138: @aviatesk do you know if it is safe to discard edges for `OptimizationState`? My understanding is that edges anyways come from `opt.inlining.edges`, so the structure already has them, but before that we used to take the `OptimizationState` -> `CodeInfo` path and attach edges to the resulting `CodeInfo`.

serenity4:
Whoops, I may have been to eager to get rid of `finish!` to keep support for 1.12+ :eyes: will add that back behind a version check.

test/cthulhu.jl
```diff
@@ -0,0 +1,3 @@
+using Cthulhu: CTHULHU_MODULE
+const Cthulhu = CTHULHU_MODULE[]
+const CC = Cthulhu.CC
```
serenity4, line 3: That's to ensure that we can do `using Compiler` and have tests switch to using the extension module and keep working.

serenity4:
Marking as draft for now, there will be upcoming refactors to related functions in Compiler.

PR #640, opened 2025-04-23
Preserve `CallMeta` results during inference

serenity4:
Store `CallMeta` information emitted as a result of call inference, so that we may use it during analysis once inference is done. It should theoretically provide more refined `rt`/`exct`/`effects` values, and it additionally contains `refinements` information should we ever want to support introspecting into that (which could prove helpful e.g. for https://github.com/JuliaLang/julia/pull/57651).

This is performed by defining `CthulhuCallInfo <: CC.CallInfo` which is used as a wrapper for (almost) all `CC.CallInfo` structures, applied right after inference and propagated throughout optimization (notably, through inlining). Special handling had to be done for `CC.UnionSplitApplyCallInfo`, but as far I as am aware this should not be limiting in any way.

We might be able to refactor and trim a few of our `CallInfo` structures with this change, I plan to tackle that next. In the current state, I think this can be merged safely to keep changes incremental and small, if we are willing to commit to this design.
I'd be happy to have any pointers on which example may bring to light the theoretical improvements of keeping `CallMeta` information for `rt`/`exct`/`effects`, so we can add a test and have this PR be a clear and net improvement overall.

Many thanks to @aviatesk for suggesting this approach and paving the way with explanations on how to integrate custom `CC.CallInfo` into the compilation pipeline.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
 name = "Cthulhu"
 uuid = "f68482b8-f384-11e8-15f7-abe071a5a75f"
-version = "2.17.2"
+version = "2.17.3"
 
 [compat]
 CodeTracking = "0.5, 1"
--- a/src/CthulhuBase.jl
+++ b/src/CthulhuBase.jl
@@ -11,7 +11,7 @@ using WidthLimitedIO
 
 using Core: MethodInstance, MethodMatch
 using Core.IR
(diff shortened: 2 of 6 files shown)


aviatesk (approved):
Sorry for the significant delay in reviewing this.
I think the overall approach and implementation look good. Thanks.
The necessary special case for `UnionSplitApplyCallInfo` is a bit of a pain, but I think it's okay to hold off on a fundamental fix (like preparing an interface on the inlining algorithm side) until we actually run into an issue.
LGTM.

serenity4:
No worries, I understand you've been quite busy these days :)
I 100% agree with your comment on `UnionSplitApplyCallInfo`.

PR #641, opened 2025-05-09
Adjust to `finishinfer!` signature change

serenity4:
Fixes a breakage introduced by a signature change in https://github.com/JuliaLang/julia/pull/58343.


serenity4:
The Julia PR was marked as backport for 1.12, so we'll have to fix it as well when backported. In the meantime, I'll leave things as they are to avoid breaking 1.12 for a little while.

PR #643, opened 2025-05-14
Do not rely on `Base.active_repl`

### LINKED ISSUE #450: UndefVarError: active_repl not defined in jupyter notebook
The following code errors on jupyter notebook, but works fine in Julia REPL.

```julia
using Cthulhu
function test(x)
    return x + 2
end
@descend test(1)
```

```
UndefVarError: active_repl not defined

Stacktrace:
 [1] default_terminal()
   @ Cthulhu /data.lmp/nrana/.julia/packages/Cthulhu/KIvdU/src/Cthulhu.jl:255
 [2] _descend_with_error_handling(f::Any, argtypes::Any; kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:iswarn,), Tuple{Bool}}})
   @ Cthulhu /data.lmp/nrana/.julia/packages/Cthulhu/KIvdU/src/Cthulhu.jl:233
 [3] descend_code_typed(::Any, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Cthulhu /data.lmp/nrana/.julia/packages/Cthulhu/KIvdU/src/Cthulhu.jl:178
 [4] descend_code_typed(::Any, ::Any)
   @ Cthulhu /data.lmp/nrana/.julia/packages/Cthulhu/KIvdU/src/Cthulhu.jl:178
 [5] top-level scope
   @ In[3]:5
```

`julia version 1.8.0`
`Cthulu v2.8.14`

serenity4:
Relying on `Base.active_repl` introduced two failure modes:
- If `Cthulhu.REPL !== parentmodule(typeof(Base.active_repl))`, we error.
- If we launch in non-interactive mode, `Base.active_repl` is not defined and we error.

We used it only to retrieve a terminal, but as it turns out the setup is [extremely simple](https://github.com/JuliaLang/julia/blob/2c2114ceed8a538290fbb6413e314c28bfd5b4f3/base/client.jl#L472-L473) so it seems we can just do that ourselves. From what I understand, it hooks into `stdout`/`stdin`/`stderr`, but nothing requires interacting with `Base.active_repl`.

This should fix #450.

Reproducing the first failure mode (and ensuring it is now fixed) is a bit tricky, these conditions seem to work for me:
- Copy the REPL package from its stdlib location to somewhere else (e.g. `~/.julia/dev`).
- `dev` it within the Cthulhu project.
- Launch Julia with `julia --project --startup-file=no` then execute `using Cthulhu; @descend exp(4)`.
- This should error on `master` and pass on this PR.

The second failure mode can be reproduced with https://github.com/JuliaDebug/Cthulhu.jl/issues/450#issuecomment-2877185147.

### DIFF
--- a/src/CthulhuBase.jl
+++ b/src/CthulhuBase.jl
@@ -142,7 +142,11 @@ function __descend_with_error_handling(args...; terminal=default_terminal(), kwa
     return nothing
 end
 
-default_terminal() = REPL.LineEdit.terminal(Base.active_repl)
+function default_terminal()
+    term_env = get(ENV, "TERM", @static Sys.iswindows() ? "" : "dumb")
+    term = REPL.Terminals.TTYTerminal(term_env, stdin, stdout, stderr)
+    return term
+end
 
 descend_impl(@nospecialize(args...); kwargs...) =
     _descend_with_error_handling(args...; iswarn=true, kwargs...)
--- a/test/test_terminal.jl
+++ b/test/test_terminal.jl
@@ -40,9 +40,7 @@ macro with_try_stderr(out, expr)
 end
 
 @testset "Terminal" begin
-    if isdefined(Base, :active_repl)
-        @test Cthulhu.default_terminal() isa REPL.Terminals.TTYTerminal
-    end
+    @test Cthulhu.default_terminal() isa REPL.Terminals.TTYTerminal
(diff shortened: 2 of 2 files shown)


aviatesk (approved):
Thanks, confirmed this fixes the both failure modes.

PR #645, opened 2025-05-29
Adjust to method table changes

serenity4:
`fT.name.mt.name` was replaced by `fT.name.singletonname` in https://github.com/JuliaLang/julia/pull/58131, which caused a breakage.


aviatesk (approved):
> it used to access the method table from MethodList, and AFAIU it's no longer possible to do so.

So it seems that we need to change the fundamental implementation strategy of MethodAnalysis?

PR #654, opened 2025-07-17
Re-enable (and fix) terminal tests

serenity4:
I fixed and refactored tests for `test_terminal.jl`, which now all pass except one where it seems that the Revise integration does not work for some reason. I marked it as broken so we can later address it.

I'm not sure about enabling/fixing IRShow tests right now, because it seems like they would very easily break by testing against hardcoded IR. I would favor having more precise tests that cover styling (colors etc) as part of the terminal tests, which we can add in the future.


aviatesk (approved):
Thanks so much for working on this!

serenity4:
@aviatesk feel free to review changes introduced in this additional commit: e262ead546465c57af23549c20f1d88d9b91bb01

If your approval still stands, I think we can merge this :)

aviatesk (approved):
Changes looks great. Thanks for working on this!

PR #655, opened 2025-07-24
Allow control of terminal pager via kwargs

timholy:
This accepts keywords from `FoldingTrees.TreeMenu`. The most useful
option is perhaps `ascend(obj; pagesize=20)`.

CC @fingolfin


serenity4 (approved):
Seems like a nice QoL feature, thanks @timholy!

fingolfin:
Thanks @timholy ! Can one also use this with e.g. `descend` and `@descend`, and if so: how?

timholy:
@fingolfin see #656

PR #660, opened 2025-08-06
Bump CodeTracking compat

oscardssmith:
https://github.com/JuliaDebug/Cthulhu.jl/pull/658 did this for Cthulhu, but forgot to update CodeTracking


serenity4:
Ah, right, we were missing that for TypedSyntax. Thanks :+1:

PR #672, opened 2025-09-30
lookup: fix semiconcrete calls with annotate_types=true

aviatesk:
Currently an error occurs when descending into semi-concrete interpreted calls with `annotate_types=true`. This happens because the current implementation of `lookup_semiconcrete` does not handle the `!optimize` case at all.

This commit resolves the issue by passing `ci::CodeInstance` to `lookup_semiconcrete`, allowing it to call `lookup()` from there.

MRE:
```julia
julia> callsincos(x, y) = sin(x), cos(y);

julia> descend((Int,); annotate_types=true) do x
           callsincos(x, 42.)
       end # then descend into `callsincos`
```

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -31,8 +31,8 @@ lookup_semiconcrete(interp::AbstractInterpreter, curs::AbstractCursor, override:
 missing `$AbstractCursor` API:
 `$(typeof(curs))` is required to implement the `$lookup_semicocnrete(interp::$(typeof(interp)), curs::$(typeof(curs)), override::SemiConcreteCallInfo, optimize::Bool)` interface.
 """)
-lookup_semiconcrete(interp::CthulhuInterpreter, ::CthulhuCursor, override::SemiConcreteCallInfo, optimize::Bool) =
-    lookup_semiconcrete(interp, override, optimize)
+lookup_semiconcrete(interp::CthulhuInterpreter, curs::CthulhuCursor, override::SemiConcreteCallInfo, optimize::Bool) =
+    lookup_semiconcrete(interp, get_ci(curs), override, optimize)
 
 get_ci(curs::AbstractCursor) = error(lazy"""
 missing `$AbstractCursor` API:
--- a/src/CthulhuBase.jl
+++ b/src/CthulhuBase.jl
@@ -264,14 +264,14 @@ function lookup_constproped_unoptimized(interp::CthulhuInterpreter, override::In
     return LookupResult(src, rt, exct, infos, slottypes, effects, codeinf)
 end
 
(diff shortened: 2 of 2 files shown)


src/interface.jl
```diff
@@ -31,8 +31,8 @@ lookup_semiconcrete(interp::AbstractInterpreter, curs::AbstractCursor, override:
 missing `$AbstractCursor` API:
 `$(typeof(curs))` is required to implement the `$lookup_semicocnrete(interp::$(typeof(interp)), curs::$(typeof(curs)), override::SemiConcreteCallInfo, optimize::Bool)` interface.
 """)
-lookup_semiconcrete(interp::CthulhuInterpreter, ::CthulhuCursor, override::SemiConcreteCallInfo, optimize::Bool) =
-    lookup_semiconcrete(interp, override, optimize)
+lookup_semiconcrete(interp::CthulhuInterpreter, curs::CthulhuCursor, override::SemiConcreteCallInfo, optimize::Bool) =
+    lookup_semiconcrete(interp, get_ci(curs), override, optimize)
```
serenity4, line 35: Isn't this `CodeInstance` that of the caller?

In #662 I went with converting the `IRCode` we get to a `CodeInfo` on the spot.

serenity4: > I think get_ci(curs) is supposed to be the code instance used for regular inference for this call (since we also use this pattern for the general lookup interface)?

Quite possibly :thinking: if you tested it and it works, I'd say let's go with that :)

> Yeah, that sounds the right approach.

I'd say so only if the `CodeInstance` is less precise than the `IRCode`. I am a bit puzzled by the fact that we emit a `CodeInstance` for something that we have an `IRCode` for and that we don't intend to cache (AFAIK), but I may not have all the pieces.

aviatesk: > I'd say so only if the `CodeInstance` is less precise than the `IRCode`. I am a bit puzzled by the fact that we emit a `CodeInstance` for something that we have an `IRCode` for and that we don't intend to cache (AFAIK), but I may not have all the pieces.

We do not prepare new `CodeInstance` when performing local-only inference (e.g. const-prop, concrete-eval, semi-concrete interpret), but I limited conditions like call-site inlining and external abstract interpreters may do so.

Moving back to this code, the current implementation of `annotate_types` requires `CodeInfo`, but in the case of semi-concrete interpret, there is essentially no corresponding `CodeInfo` and only `IRCode` exists, which causes problems. In this commit, we pass the `CodeInfo` from the regular inference results before semi-concrete interpret to `annotate_types`, but it would be more correct to convert the `IRCode` inferred by semi-concrete interpret to `CodeInfo`.

serenity4: Ah I see the full picture now, thanks for the explanations! :pray:

aviatesk: You're welcome! Sorry to make conflicts to your PR.

PR #677, opened 2025-11-04
Split compiler integration to avoid duplicating Cthulhu when loading the compiler extension

serenity4:
This PR splits the compiler integration functionality, such that we only need to re-include the corresponding few compiler-related files when we load the Compiler extension, instead of re-loading most of Cthulhu. These files live under `src/compiler/*.jl` and essentially hold the interpreter and default provider bits (see `src/CthulhuCompiler.jl` which includes these).

A major benefit of this approach is that we now have a single API regardless of whether to use the compiler extension: most user-facing data structures live in Cthulhu, and only the parts related to the interfacing with the compiler are either in Cthulhu or CthulhuCompilerExt. No more `Cthulhu.AbstractProvider !== CthulhuCompilerExt.AbstractProvider`. However, `Cthulhu.DefaultProvider !== CthulhuCompilerExt.DefaultProvider` may still be true. When we load the Compiler extension, we fall in one of two situations:
- The compiler stdlib is just a reexport of `Base.Compiler`, in which case we don't need to do anything (`Base.Compiler.AbstractInterpreter === Compiler.AbstractInterpreter`).
- The compiler stdlib defines data structures that are different than those from `Base.Compiler`, so we re-include `src/CthulhuCompiler.jl` which defines new structs and functions, and adds methods to Cthulhu functions.

Another benefit that comes from this is that we won't pay the cost of recompilation in the most common case, i.e. Compiler may be loaded but simply aliases `Base.Compiler`, as the Cthulhu extension is virtually a no-op in this situation.


aviatesk:
Sorry for the delay in reviewing this. After reading the PR description, I think it's a great change overall.

There are a few things I can't understand just by reading the code, so I'd like to clarify them:
- If I load a custom `Compiler.jl` and want to use Cthulhu with it, how should I do that? With the previous `CthulhuCompilerExt` implementation, I believe that when a custom `Compiler.jl` was loaded, the `descend` implementation would automatically switch, and `descend` would then work with that new compiler. From what I can tell by following the current implementation, this doesn't seem to happen, but am I missing something? I don't necessarily think it should happen automatically, but I'd like to confirm the steps needed to use `descend` with a new compiler.
- If an external `AbstractInterpreter` package wants to use Cthulhu's `AbstractProvider` interface, my understanding is that it should primarily use `Cthulhu.get_module_for_compiler_integration(; use_compiler_stdlib = true)` to get the Cthulhu module to overload. Is that correct? And my understanding is that this Cthulhu module will be correctly switched depending on the `Compiler.jl` implementation.

aviatesk:
> If I load a custom Compiler.jl and want to use Cthulhu with it, how should I do that?

Looks like I can just do
```julia-repl
julia> using Cthulhu

julia> using Compiler # load custom Compiler impl

julia> descend(sin, (Int,); interp=Compiler.NativeInterpreter())
```
and it's using the new Compiler implementation. That's impressive.

aviatesk (approved):
> I'll add a few more docs to provide guidelines and explanations about how to do all that integration before merging, hopefully this PR simplifies the whole process.

That's great.

I haven't completed the JET integration with `AbstractProvider` on my side yet, but I'll try it again on this branch. Last time I tried, I remember running into an error related to this Compiler extension complexity, but looking at the changes in this PR, it seems that's been fixed entirely. (the challenge of integrating with JET is actually the other thing, specifically, I need to properly manage `CodeInfo` in `JET.AbstractAnalyzer` for passing them to `AbstractProvider`. But after this PR, once that's done, Cthulhu should be usable with JET.)

aviatesk:
Feel free to merge this whenever you are satisfied.

serenity4:
> Feel free to merge this whenever you are satisfied.

That should be good now, I'll proceed with the merge and tagging of the breaking release 3.0.0 with adequate release notes.

PR #678, opened 2025-11-20
Fix formatting and typos in CONFIG docs

wheeheee:
Looks weird in Julia's help mode, so I tried to escape hide_type_stable's underlining (which should be italic in Markdown?).


aviatesk:
Thanks!

PR #684, opened 2025-12-04
CI: Update Julia versions

aviatesk:

PR #687, opened 2025-12-04
improve `AsyncIO` robustness

aviatesk:
~~- Fix `unsafe_read` to correctly calculate remaining bytes (`nb - written`) and use proper destination offset (`to + written`)~~
~~- Fix potential race condition in `peek` by checking buffer under lock~~

Written with Claude.

EDIT: Turns out I just ended up crafting noises. Picked up one minor improvement from it.

### DIFF
--- a/src/testing.jl
+++ b/src/testing.jl
@@ -109,7 +109,7 @@ function Base.unsafe_read(io::AsyncIO, to::Ptr{UInt8}, nb::UInt)
             splice!(buffer, 1:n)
             written += n
         end
-        eof(io) && throw(EOFError())
+        istaskdone(io.task) && throw(EOFError())
     end
 end
 
--- a/src/testing.jl
+++ b/src/testing.jl
@@ -73,12 +73,8 @@ end
 AsyncIO(terminal::VirtualTerminal) = AsyncIO(terminal.output)
 
 function Base.peek(io::AsyncIO, ::Type{UInt8})
-    while true
-        @lock io.lock begin
-            isempty(io.buffer) || return io.buffer[1]
-        end
-        yield()
-    end
+    while bytesavailable(io) < 1 yield() end
+    return @lock io.lock io.buffer[1]
 end
 function Base.read(io::AsyncIO, ::Type{UInt8})
     ref = Ref(0x00)
@@ -101,24 +97,17 @@ function Base.readavailable(io::AsyncIO)
 end
 
 function Base.unsafe_read(io::AsyncIO, to::Ptr{UInt8}, nb::UInt)
-    written = UInt(0)
+    written = 0
     while written < nb
-        n = @lock io.lock begin
+        bytesavailable(io) > 0 || (yield(); continue)
+        @lock io.lock begin
             (; buffer) = io
-            n = min(UInt(length(buffer)), nb - written)
-            if n > 0
-                GC.@preserve buffer begin
-                    unsafe_copyto!(to + written, pointer(buffer), n)
-                end
-                splice!(buffer, 1:n)
+            n = min(length(buffer), nb)
+            GC.@preserve buffer begin
+                unsafe_copyto!(to, pointer(buffer), n)
             end
-            n
-        end
-        written += n
-        if written < nb
-            # Yield first to give the background task a chance to read more data
-            yield()
-            eof(io) && throw(EOFError())
+            splice!(buffer, 1:n)
+            written += n
         end
     end
 end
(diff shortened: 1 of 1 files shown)


src/testing.jl
```diff
@@ -73,8 +73,12 @@ end
...
 
 function Base.peek(io::AsyncIO, ::Type{UInt8})
-    while bytesavailable(io) < 1 yield() end
-    return @lock io.lock io.buffer[1]
+    while true
+        @lock io.lock begin
+            isempty(io.buffer) || return io.buffer[1]
+        end
+        yield()
+    end
```
serenity4: The reason reported on the PR description was
> Fix potential race condition in peek by checking buffer under lock

But a simple look at that portion of code should have already revealed that there is no race condition here since `bytesavailable` already acquires the lock and uses `length(io.buffer)`.

These changes do not modify the behavior of the function, it appears to be a simple rewrite which I don't think brings any more clarity, on the contrary.

src/testing.jl
```diff
@@ -97,17 +101,24 @@ function Base.readavailable(io::AsyncIO)
...
             end
-            splice!(buffer, 1:n)
-            written += n
+            n
+        end
+        written += n
+        if written < nb
+            # Yield first to give the background task a chance to read more data
+            yield()
+            eof(io) && throw(EOFError())
```
serenity4: From what I can see, on a semantic level, the changes:
- perform multiple `unsafe_copyto!` instead of waiting for the buffer to have all the data then copy in one go.
- check for `eof(io)` at the end of every iteration.

But nothing related to
> Fix unsafe_read to correctly calculate remaining bytes (nb - written) and use proper destination offset (to + written)

and we should already correctly calculate the right number of bytes.

The rest seems to be mostly noise. If the issue was that we need an extra `istaskdone(io.task) && throw(EOFError())` in the loop, I'm all good with it, but through the LLM noise that single line appears to be the only change worth noting and the rest just increases the complexity for no reason AFAIU.

serenity4 (changes requested):
This seems to be 95% LLM noise. It was a bit hard to understand what it's supposed to do and why. Does it address the issues that you are observing? It would be nice to have a link to the original error to see what needs fixing/what to look for.

The addition on line 121 of checking that the task is still running in `unsafe_read` seems worth adding, but I would oppose the rest of the changes as it is.

aviatesk:
Thanks for the review. Looks like I submitted a low-quality PR because I didn't properly review Claude's output. Sorry for that. I made these changes while debugging #684.

Actually, `test_provider.jl` still fails on my local machine (especially with remarks not emitting correctly for some reason). But since it passes on CI, I guess it might be a non-determinism issue from `AsyncIO`, and that why I asked Claude for investigating the potential issues of this code. But it could just be a problem with my machine, though.

About this PR, you're right, I wrote it randomly. I just thought it would be a straightforward fix and created this PR taking what Claude saying, but I should have reviewed these changes more carefully..

src/testing.jl
aviatesk: I think Claude was probably worried about a scenario where there's a write to `io` between this `while` loop and `return`. But practically, I don't think it'll be an issue.

src/testing.jl
```diff
@@ -109,6 +109,7 @@ function Base.unsafe_read(io::AsyncIO, to::Ptr{UInt8}, nb::UInt)
             splice!(buffer, 1:n)
             written += n
         end
+        eof(io) && throw(EOFError())
```
serenity4:
```suggestion
        istaskdone(io.task) && throw(EOFError())
```

The second part of `eof` checks `@lock io.lock isempty(io.buffer)`, which may be false if we are waiting for new bytes to be written before copying the whole `nb` bytes at once.

serenity4:
No harm done! Apologies if I may have sounded a bit salty in my comments.

I ran the tests on my machine with Julia 1.12.2, and they all pass without any issues. If you ever happen to have a reproducer, or just see it fail on CI, I can take a look at potential issues. The whole point of this `AsyncIO` wrapper was to avoid hangs or other annoyances due to having IO interactions across tasks, so if it somehow causes weird bugs, that should definitely be fixed.

aviatesk:
> No harm

No, I totally get that it's annoying to review random code, whether it's AI-generated or not, so I'm just sorry I put you through that.
About those failing tests on my machine, they're still happening. But since they're not failing in CI, it might be something with my local preference settings. Anyway, if I can reproduce it on CI or in a whatever sharable way, I'll let you know.

PR #707, opened 2026-08-20
Make get_effects and get_exct shared generics

jishnub:
`src/CthulhuCompiler.jl` imports the `CallInfo` accessors from `Cthulhu` so that both compiler integrations extend the same generic functions — but `get_effects` and `get_exct` are missing from that import list, and `get_effects` is instead defined fresh at the bottom of the file.

Since `CthulhuCompiler.jl` is included once into `Cthulhu` and again into `CthulhuCompilerExt` when `Compiler.AbstractInterpreter !== Base.Compiler.AbstractInterpreter`, this means `Cthulhu.get_effects` and `CthulhuCompilerExt.get_effects` become two distinct functions in that configuration, and `Cthulhu.get_effects(info)` raises a `MethodError` for any `CallInfo` produced by the extension. `get_ci`, `get_rt` and the rest are unaffected because they are imported.

This adds `function get_effects end` / `function get_exct end` next to the existing stubs in `callsite.jl` (they had no owning declaration before) and adds both names to the import list.

No behaviour change in the common case, where `Compiler` reexports `Base.Compiler` and the extension is a no-op.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


serenity4:
Makes sense, thanks!

